PR #13, opened 2016-08-20
bundle builds (types & js) to es5

### LINKED ISSUE #12: Make ES5-compatible build

Speedy37:
Fixes #12


Speedy37:
sry, I messed a bit my PR, 
the second commit is about exporting typings, so there is no need to do npm install @types/**

joebehymer:
+1.  We're using TS ES5 and would love to use this.

bcherny:
Thanks for the contribution @Speedy37! Can you help me understand what /out and /dist are for? If you'd like to merge this more quickly, feel free to cherry pick dd9519e9ceb7e56cd30ab9b026109ac9f1813a49 and create a PR with just that commit.

Speedy37:
out/ is the temporary output directory where compiled .ts & .ts.d files are (made by tsc)
dist/ is the bundled .js and .ts.d files (made with dts-bundle & browserify)

bcherny:
Can we use tsify instead to avoid temporary files?

Speedy37:
we can, but types declaration would be missing.

bcherny:
I see what you're saying. Let's merge this guy, and I'll do some cleanup later.

PR #18, opened 2016-09-06
Fixes #17: title attribute used as type and does not resolve

### LINKED ISSUE #17: title attribute used as type and does not resolve
I noticed that any property in a json schema which has a title does not get resolved. Instead, the title is used as a type.

Example:

        "users": {
          "type": "array",
          "title": "user id array",
          "description": "Array of authorized user ids.",
          "items": {
            "type": "string"
          }
        }

gets translated to:

    users?: UserIdArray;

which leads to a undefined type `UserIdArray`. Is that behavior expected? How do I get json-schema-to-typescript to define the referenced types as well?

For now I created a workaround by removing ` || rule.title` from [line 168](https://github.com/bcherny/json-schema-to-typescript/blob/master/src/index.ts#L168). This generates:

    users?: string[];

Speedy37:


bcherny:
Fantastic, thanks @Speedy37!

Issue #24, opened 2016-09-19
Description with new lines `\n` are not output correctly in typescript comments


darcyparker:
If I have a description like this in my JSON Schema:

```JSON
"description": "Some description\nWith multiple lines"
```

The output typescript definition neglects to comment the new lines too.

It would be great if long term, the description comment could support multiple lines. For the short term, I am simply stripping the new lines out of the JSON schema before I compile it to typescript.

bcherny:
Not quite a duplicate of, but very similar to https://github.com/bcherny/json-schema-to-typescript/issues/10. Feel free to put up a PR, otherwise I will get to this eventually.

darcyparker:
See https://github.com/bcherny/json-schema-to-typescript/pull/25

darcyparker:
Please see new pull request: https://github.com/bcherny/json-schema-to-typescript/pull/26

donaldpipowitch:
FYI: It would be nice to use `/** */` block comments for VS Code. See https://github.com/bcherny/json-schema-to-typescript/pull/26#discussion_r79550155.

PR #26, opened 2016-09-19
Multiple line description support

darcyparker:
Pull request to fix https://github.com/bcherny/json-schema-to-typescript/issues/24

This replaces my previous pull request that concatenated multiple line descriptions.

### DIFF
--- a/src/TsTypes.ts
+++ b/src/TsTypes.ts
@@ -1,5 +1,10 @@
 import { camelCase, upperFirst } from 'lodash'
 
+const multiLineCommentStart  = '/** '
+const multiLineCommentIndent = ' *  '
+const multiLineCommentEnd    = ' */'
+const newLineRegex = /\\n|\n/
+
 export namespace TsType {
 
   export interface TsTypeSettings {
@@ -36,7 +41,13 @@ export namespace TsType {
       return this.id && upperFirst(camelCase(this.id))
     }
     protected toBlockComment(settings: TsTypeSettings) {
-      return this.description && settings.declarationDescription ? `/** ${this.description} */\n` : ''
+      return this.description && settings.declarationDescription ?
+        `${this.description
+            .split(newLineRegex)
+            .map((line, lineNum) => (lineNum > 0 ? multiLineCommentIndent : multiLineCommentStart) + line)
+            .join('\n') + multiLineCommentEnd + '\n'
+        }` :
+        ''
     }
     protected _toDeclaration(decl: string, settings: TsTypeSettings): string {
       return this.toBlockComment(settings) + decl + (settings.endTypeWithSemicolon ? ';' : '')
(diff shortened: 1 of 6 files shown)


test/cases/with-description-newlines.ts
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "required": ["firstName", "lastName"]
+}
+
+export var types = `/** My cool schema */
+export interface ExampleSchema {
+  firstName: string; // first name single line description
+  lastName: string;
+  // Age description with
+  // multiple lines
```
bcherny: Possible to use a multiline comment instead?

darcyparker: Done in https://github.com/bcherny/json-schema-to-typescript/pull/26/commits/adace03e6f19311050885f91185ab39ad5caab45

dist/index.js
```diff
@@ -233,7 +236,10 @@ var TsType;
                 //Multiple line description in comment above
                 if (settings.propertyDescription && _.type.description && !_.type.id)
                     decl = (_.type.description.search(newLineRegex) > 0) ?
-                        _.type.description.split(newLineRegex).map(function (line) { return indentString + '// ' + line; }).join('\n') + '\n' + decl :
+                        _.type.description
+                            .split(newLineRegex)
+                            .map(function (line, lineNum) { return indentString + (lineNum > 0 ? multiLineCommentIndent : multiLineCommentStart) + line; })
```
darcyparker: As you know, `index.js` is generated by `gulp dist`. Curiously, it is declaring `function (line, lineNum)` rather than just `(line, lineNum)`...

src/TsTypes.ts
```diff
@@ -1,5 +1,9 @@
 import { camelCase, upperFirst } from 'lodash'
 
+const multiLineCommentStart =  '/** '
+const multiLineCommentIndent = '    '
```
darcyparker: BTW: If you like multi line comments like this:
```
/**
  *
  *
*/
```

its a simple change... Just define: `const multiLineCommentIndent = '  * '`

bcherny: Bummer- I was trying to find a comment syntax that VSCode will use to inform its intellisense, but it doesn't look like comments for interfaces are supported.

donaldpipowitch: I just tried to create a new issue to ask for a feature request to only use jsdoc comments for `descriptions`, before I saw this...

I don't know how you tried it, but VSCode works wonderfully if you use `/** */` in interfaces. It would be _really_ useful, if we could move descriptions to `/** */` comments.
image

darcyparker: I am using vim with https://github.com/Quramy/tsuquyomi and some other typescript plugins to get similar code completion, hints, jump to definition/references, etc..  Under the covers, the plugin uses a service from `tsc` and I guess VSCode uses this same service.

I found the hinting works when the interface is in the same file. But if it is imported, the hint points to the import. I have to jump to the definition to see the comment.  Perhaps this is the difference in @donaldpipowitch and @bcherny test?  Maybe one day `tsc` will be enhanced to provide hints to interfaces that are external to the current file?

@bcherny - what do you think? Should I retire the code that does the single line description?  I agree with @donaldpipowitch that it is really useful to have this info available to editors.

src/TsTypes.ts
```diff
@@ -181,14 +185,26 @@ export namespace TsType {
...
             decl += '?'
           decl += ': ' + _.type.toType(settings)
           if (settings.endPropertyWithSemicolon)
             decl += ';'
+
+          //Single line description in comment to the right of declaration
+          //Multiple line description in comment above
           if (settings.propertyDescription && _.type.description && !_.type.id)
-            decl += ' // ' + _.type.description
+            decl = (_.type.description.search(newLineRegex) > 0) ?
```
darcyparker: Good catch. I updated to use `test()`.

darcyparker:
Please see https://github.com/bcherny/json-schema-to-typescript/pull/26/commits/0d5466d4f23fba17a7b08fb463111db2289646e1 I rebased ontop of the latest `master` branch and squashed because the rebase changed the commit ids.  @bcherny - your feedback about using `test()` is incorporated.

src/TsTypes.ts
bcherny: Ah, different use cases. When you're typing out `bar`, it should the description. But after you've entered `bar: someValue` and hover over `bar` again, the description is lost.

Anyway, @darcyparker I agree - if you want to tackle it in a separate commit in this PR, feel free to get rid of single line comments in favor of multiline jsdoc-style comments for intellisense. Otherwise, feel free to do it in a separate PR, or just file an issue.

darcyparker: @bcherny - Done. I put it in same pull request, but two separate commits. I also rebased on top of latest `master` branch. See https://github.com/bcherny/json-schema-to-typescript/pull/26/commits/aef14ff462a106b76eadd0be8babc08e26fec14b

donaldpipowitch:
This is great. Thank you.

PR #30, opened 2016-09-20
Update npm scripts to use symlinked bin dir

darcyparker:
This is a minor change so that scripts run better on all OSes (windows, unix flavors).  The scripts in `npm_modules/.bin/*` will be the correct version for the project and the user's OS (windows or unix flavor).

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,11 +5,12 @@
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
   "scripts": {
-    "tsc": "tsc",
-    "build": "gulp build",
-    "test": "gulp test && gulp lint",
-    "debug": "gulp debug",
-    "watch": "watch \"npm run test\" ./src/ ./test/",
+    "tsc": "./node_modules/.bin/tsc",
+    "gulp": "./node_modules/.bin/gulp",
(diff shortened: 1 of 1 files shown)


darcyparker:
Note: I am not sure https://github.com/bcherny/json-schema-to-typescript/pull/30/commits/bb6c5bc1f551c4ce839b8d82de5f847ff8dc0f67#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R13 is optimal because the `path` arguments are unix flavor.  And similarly https://github.com/bcherny/json-schema-to-typescript/pull/30/commits/bb6c5bc1f551c4ce839b8d82de5f847ff8dc0f67#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R11 has `&&` which is for posix shells like `bash`.

It might be best to have an OS agnostic `.js` script.

bcherny:
NPM already prefers scripts in .bin, see https://docs.npmjs.com/misc/scripts#path. Not sure I understand why this PR is necessary.

darcyparker:
I didn't know about the `path`. Thanks for the reference. I must have been confused. I was getting errors about not having `gulp` installed. I'll close this.

PR #31, opened 2016-09-20
Fix minor tslint issues

darcyparker:


test/cases/enumValidation.2.ts
```diff
@@ -17,6 +17,6 @@ export var settings = {
 }
 
 export var error = {
-  type: TypeError,
-  message: 'Enum was declared as an integer type, but found at least one non-string tsEnumValue'
+  message: 'Enum was declared as an integer type, but found at least one non-string tsEnumValue',
```
darcyparker: Sorted order of keys... https://github.com/bcherny/json-schema-to-typescript/blob/master/tslint.json#L25

PR #36, opened 2016-09-21
Minor ts lint fixes

darcyparker:
I was getting and ignoring some tslint issues about a few cases and fixed them here.

When reviewing, please look at https://github.com/bcherny/json-schema-to-typescript/pull/36/commits/150110e18d5cccb201eeaa37b7668cba96f46544#diff-f41e9d04a45c83f3b6f6e630f10117feR171 carefully.  I believe it is correct, but clearly a bit tricker.

No hurry to review this. Take your time.

(I also removed an unnecessary `lines` argument that I should have caught in a previous pull request.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "json-schema-to-typescript",
-  "version": "2.1.2",
+  "version": "2.2.0",
   "description": "compile json schema to typescript typings",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
--- a/src/TsTypes.ts
+++ b/src/TsTypes.ts
@@ -1,5 +1,10 @@
 import { camelCase, upperFirst } from 'lodash'
 
+const multiLineCommentStart  = '/** '
+const multiLineCommentIndent = ' *  '
+const multiLineCommentEnd    = ' */'
+const newLineRegex = /\\n|\n/
+
 export namespace TsType {
 
   export interface TsTypeSettings {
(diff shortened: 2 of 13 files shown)


src/index.ts
```diff
@@ -168,7 +168,11 @@ class Compiler {
     switch (typeof a) {
       case 'boolean': return new TsType.Boolean // ts doesn't support literal boolean types
       case 'number': return new TsType.Number   // ts doesn't support literal numeric types
-      case 'string': return new TsType.Literal(JSON.stringify(a))
+      case 'string':
+        // output single quotes for tslint with inner quotes (if any) escapped
+        var stringValue = JSON.stringify(a)
+        stringValue = `'${stringValue.substring(1, stringValue.length - 1).replace(/'/g,'\\\'')}'`
+        return new TsType.Literal(stringValue)
```
bcherny: Let's do this in a separate PR, with tests please. The rest looks good!

darcyparker: Tests are a good idea. I can do that. This is connected to some of the changes though.

For example:
* https://github.com/bcherny/json-schema-to-typescript/pull/36/commits/150110e18d5cccb201eeaa37b7668cba96f46544#diff-22946053a913fc1293f4df407a6992f4R44 changes the double quote to single quote. And the above code change in index.ts is required to validate successfully.
* There are many other locations of the double quotes too.

So I will break this up into 2 pull requests. One just for the double quote tslint issues and the other for all remaining tslint issues.

Issue #34, opened 2016-09-21
type of enums should not default to string


donaldpipowitch:
See https://github.com/bcherny/json-schema-to-typescript/blob/master/src/index.ts#L251.

As far as I understand the standard it is allowed to create a schema like `{ enum: [ false ] }`. Currently an error is thrown if you use `{ enum: [ false ] }`: `TypeError: Enum was declared as a string type but found at least one non-string member`.

E.g. given

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "foo": {
      "enum": [ false ]
    }
  }
}
```

and 

```
{
  "foo": false
}
```

is valid here: http://www.jsonschemavalidator.net/.

["understanding json schema"](https://spacetelescope.github.io/understanding-json-schema/reference/generic.html#enumerated-values) also says "You can use enum even without a type, to accept values of different types."

---

Okay, I see I can't even workaround this by using `{ type: 'boolean', enum: [ false ] }`, because of https://github.com/bcherny/json-schema-to-typescript/blob/master/src/index.ts#L257: `TypeError: Enum type must be string or integer; default is string if undefined`.

--- 

It is also valid to just use `false` as a type in TypeScript. See [this example](https://blogs.msdn.microsoft.com/typescript/2016/08/30/announcing-typescript-2-0-rc/):

```typescript
interface Success<T> {
    success: true;  // type is `true`, not `boolean`
    value: T;
}

interface Failure {
    success: false;  // type is `false`, not `boolean`
    reason: string;
}

type Result<T> = Success<T> | Failure;
```

bcherny:
Fixed in 3773b2477c7ada189e5e476cf9bb305b320bd150.

I'm not sure that `foo: false` is valid - can you point me to some reference material?

donaldpipowitch:
What do you mean? Well you can take the schema and payload given in the first post and try it in an online validator.

bcherny:
```json
{
  "foo": false
}
```

Entering ^ in the validator you linked says that any value of `foo` is valid.

donaldpipowitch:
I can't confirm that. For me, it looks like _only_ `false` is valid and everything else is invalid:

![image](https://cloud.githubusercontent.com/assets/1152805/18824331/9323a6e8-83c1-11e6-8437-2888d46952df.png)

[...]

PR #37, opened 2016-09-21
Support for enum of boolean

darcyparker:
See https://github.com/bcherny/json-schema-to-typescript/issues/34


src/index.ts
```diff
@@ -248,21 +251,23 @@ class Compiler {
   }
 
   private validateEnumMembers(rule: JSONSchema.Schema): EnumType {
-    if (!rule.type) rule.type = 'string'
+    if (rule.type === undefined) rule.type = 'string'
```
darcyparker: Note: `!rule.type` returns `false` if `rule.type === 'boolean'` or any string value. I think the correct intent is `rule.type === undefined`

darcyparker:
I took a crack at this because I thought the case of a type that is a subset of `boolean` was useful after reading @donaldpipowitch example.  I hope I interpreted the problem correctly. I am open to feedback or scrapping this I misunderstood the problem.

src/index.ts
darcyparker: Note: `!rule.type` returns `false` if `rule.type === 'boolean'` or any string value. I think the correct intent is `rule.type === undefined`

donaldpipowitch: I don't think this is correct. It is valid to _not_ set a type here. If we default to `'string'` and our enum is `[ false ]`, this would be incorrect.

test/cases/enumBoolean.ts
```diff
@@ -0,0 +1,23 @@
+export var schema = {
+  "title": "EnumBoolean",
+  "type": "object",
+  "properties": {
+      "foo": {
+        "type" : "boolean",
```
donaldpipowitch: Maybe we could have another test _without_ a `"type": "boolean"`. Like here: https://github.com/webpack/webpack/blob/master/schemas/webpackOptionsSchema.json#L638

src/index.ts
darcyparker: if `"type" : "boolean"`,  it will be defined, so `rule.type` won't default to `string`.

test/cases/enumBoolean.ts
darcyparker: Sure - that makes sense.  The code will break without it as it stands now. But I think it can be refactored to work.

src/index.ts
donaldpipowitch: Yes, _if_ `"type" : "boolean"` is defined. But if it isn't defined, it should _not_ default to string. This is how JSON schema works as far as I know.

darcyparker:
@donaldpipowitch and @bcherny: I refactored based on feedback and rebased. Let me know what you think now.

Note: As I started to add the new enumValidation tests cases, I found it confusing to navigate.  So I numbered them to correspond to the 6 validations in the order they are defined in validateEnumMembers(). The 3rd validation has 3 cases, so they are numbered `3.1`, `3.2`, `3.3`.

darcyparker:
Also note: I added `.nvmrc` to make it convenient to use the same node version that you are using for integration tests.  I use `avn` with `nvm` so that when I `cd` into a directory, my environment is using the right version of node.

darcyparker:
FYI: Just rebased on top of latest master branch.

test/cases/enum.ts
```diff
@@ -2,46 +2,58 @@ export var schema = {
...
+      '}',
+      'export interface Enum {',
+      '  aStringEnum: "a" | "b" | "c";',
+      '  anImpliedStringEnum: "a" | "b" | "c";',
+      '  aBooleanEnum: true;', // Note, type is `true` and not boolean (the enumeration disallows `false`)
+      '  anImpliedBooleanType: true;', // Note, type is `true` and not boolean (the enumeration disallows `false`)
+      '  anIntegerEnum: AnIntegerEnum;',
+      '  anImpliedIntegerEnum: AnImpliedIntegerEnum;',
+      '}'
+      ].join('\n')
```
bcherny: Can you just use a single template string?

darcyparker:
Thanks for accepting this @bcherny.  And thanks for taking care of the single template string. (I thought the indentation looked better... but it's just a styling opinion. I am fine with adapting.)

BTW: I am glad to see your repo improving.  It won't be long before I am using it my projects.  It's going to really help reduce the maintenance of and improve the quality of my front end to back end interfaces/coupling.

PR #50, opened 2016-10-10
adds a sample app with full configuration showing how to use it

rcavalcanti:
I'm new to the npm toolchain and had to search for a while to understand how this library should work.

I build a small sample app to play around until I get it working. 

This PR includes that sample app as a sub-folder so others can use it as example.

Moreover, it also fixes the README file. The usage example use the `await` keyword, but it's only legal inside an `async function`.


sample/foo-typescript/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "json-typescript",
```
bcherny: "json-schema-to-typescript-example"

rcavalcanti:
Hi @bcherny, 

Thanks for reviewing. Just pushed the changes you requested.

rcavalcanti:
@bcherny I did on more commit. I realized that I have added a 'Bar' and 'Baz' examples. It was  not my intention to push them.

I also changed the schema type from Foo to Person which is more in alignment with its fields

README.md
```diff
@@ -55,9 +55,17 @@ export interface ExampleSchema {
 
 ## Usage
 
+Check [sample app](sample/foo-typescript) for more details.
```
bcherny: Update path

README.md
```diff
@@ -55,9 +55,17 @@ export interface ExampleSchema {
...
+
 ```js
 import {compileFromFile} from 'json-schema-to-typescript'
 fs.writeFileSync('foo.d.ts', await compileFromFile('foo.json'))
+
+async function generate() {
+  fs.writeFileSync('dist/foo.d.ts', await compileFromFile('src/foo.json'))
+}
```
bcherny: Rm duplicate block

README.md
```diff
@@ -55,9 +55,17 @@ export interface ExampleSchema {
...
+
+generate()
```
bcherny: rm

example/package.json
```diff
@@ -0,0 +1,21 @@
...
+  "name": "json-schema-to-typescript-example",
+  "version": "1.0.0",
+  "description": "",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
+  "scripts": {   
+    "build": "npm run clean && npm run build-sources && npm run lint && node dist/index.js",
+    "clean": "rimraf dist",
+    "build-sources": "tsc -p ./tsconfig.json",
+    "lint": "tslint -c tslint.json src/*.ts",
```
bcherny: rm lint

example/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "prepublish": "npm run build"
+  },
+  
+  "dependencies": {
+    "@types/node": "4.0.30",
+    "tslint": "3.15.1",
```
bcherny: rm

example/tslint.json
```diff
@@ -0,0 +1,38 @@
+{
```
bcherny: rm this file, no need for lint in example

README.md
rcavalcanti: the duplicated block to remove it not the async function, but the one on line 62. 

It's not possible to call await outside a async function.

bcherny: The assumption is that the containing function is async. I made it the top level fn here for brevity, as most users already know that async can't be top level.

Your complete example in /example should suffice I think.

bcherny:
Thanks for the contribution @rcavalcanti!

Issue #52, opened 2016-10-12
Types generated with full qualified schema.id url


rcavalcanti:
Currently the `schema.id` is favored over the `schema.title` when generating the type name. 

For instance,
```
{
  "id": "http://dummy.com/api/person",
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string"
    },
    "lastName": {
      "type": "string"
    }
  }
}
```
will generate 
```
export interface HttpDummyComApiPerson {
  firstName?: string;
  lastName?: string;
  [k: string]: any;
}
```

Although it helps to avoid naming collisions, it's very awkward to have types named after URLs.

I would like to suggest one of the two possible improvements:

- swap the `id` and `title` on index.ts:47
```
this.id = schema.title || schema.id || this.filePath.name || 'Interface1'
```
Which will favor the `title` over the `id`. The down side being that it may cause naming collisions.

- favor the `id`, but generate a namespace

for id = "http://dummy.com/api/person"
we could have
```
namescape DummyComApi { // http protocol dropped
  export interface Person {
    firstName?: string;
    lastName?: string;
    [k: string]: any;
  }
}
```

rcavalcanti:
Another possible solution is to have it passed in the Settings. 

Btw, settings are not available on compileFromFile method, only on compile.

bcherny:
> swap the id and title on index.ts:47

I like this idea - feel free to put up a PR.

> Btw, settings are not available on compileFromFile method, only on compile.

FYI, I am working on a complete rewrite of the compiler to make it easier to work with. Again, feel free to PR if you need this now, and I will merge it.

PR #54, opened 2016-10-12
Favor schema.title over schema.id

rcavalcanti:
This PR fixes issue #52.

It also covers referenced files


test/cases/schema-title-as-type-name.ts
```diff
@@ -0,0 +1,41 @@
+export var schema = {
+    "$schema": "http://json-schema.org/draft-03/schema",
+    "id": "http://mycompany.com/api/referencing.json",
+    "title": "Referencing",
+    "type": "object",
+    "properties": {
+        "ref": {
+            "$ref": "../../test/resources/ReferencedTypeWithId.json"
```
bcherny: No need for this 2-file test case. Just have a single file with a schema that has both a `title` and and `id`, and test that `title` takes precedence.

rcavalcanti: Done, I also aligned it with what is done in ref.js test

test/cases/schema-title-as-type-name.ts
```diff
@@ -0,0 +1,48 @@
+export var schema = {
+    "$schema": "http://json-schema.org/draft-03/schema",
+    "id": "http://mycompany.com/api/referencing.json",
+    "title": "Referencing",
+    "type": "object",
+    "properties": {
+        "ref": {
+            "$ref": "../../test/resources/ReferencedType.json"
```
bcherny: Again, no need for the $ref, I think. Since you're testing that `title` takes precedence over `id`, your test case should be as slim as possible and just test that one behavior.

Concretely, if you replace the `"ref"` property with a simple `"foo": "number"` or something like that, it would simplify your test and focus on the behavior that you're actually testing.

rcavalcanti: That is what I did in the beginning, but find out that there is two different code paths depending if you are compiling a top level definition or a referenced one.

See https://github.com/bcherny/json-schema-to-typescript/pull/54/commits/833a0a11532137d79ecb5d171976fc343404bf9f#diff-f41e9d04a45c83f3b6f6e630f10117feL259

That's why I explicitly added the reference in the test case.

rcavalcanti: BTW, in the beginning I also thought that this could be tweaked with the `Settings.useFullReferencePathAsName` flag, but apparently this is not the case.

PR #58, opened 2016-10-27
add basic cli program wrapping the library

erykpiast:


bcherny:
Thanks for the contribution @erykpiast! How is this meant to be used? `cat foo.json | json2ts > out.ts`? Please add some docs if this is the case.

erykpiast:
It's not only stdin/stdout, but parameters and named parameters too. I've updated `README`.

README.md
```diff
@@ -63,6 +63,19 @@ fs.writeFileSync('dist/foo.d.ts', await compileFromFile('src/foo.json'))
 
 See [/example](example) for a fully working demo.
 
+### CLI
+Simple CLI utility is provided within the package.
+```
+cat foo.json | json2ts > foo.d.ts
+# or
+json2ts foo.json foo.d.ts
```
bcherny: Can we default `json2ts foo.json` to output foo.d.ts in the same directory as foo.json?

src/cli.ts
```diff
@@ -0,0 +1,60 @@
...
+})
+const argIn: string = argv._[0] || argv['input']
+const argOut: string = argv._[1] || argv['output']
+
+function readInput(): Promise<string> {
+  if (!argIn) {
+    return new Promise((resolve, reject) => {
+      stdin((str: string) => {
+        try {
+          resolve(str)
```
bcherny: When would this throw?

src/cli.ts
```diff
@@ -0,0 +1,60 @@
...
+    })
+  }
+
+  return readFile(argIn)
+}
+
+function writeOutput(compiled: string): Promise<void> {
+  if (!argOut) {
+    try {
+      process.stdout.write(compiled)
```
bcherny: When would this throw?

src/cli.ts
```diff
@@ -0,0 +1,60 @@
...
+const argv: minimist.ParsedArgs = minimist(process.argv.slice(2), {
+  alias: {
+    input: [ 'i' ],
+    output: [ 'o' ]
+  }
+})
+const argIn: string = argv._[0] || argv['input']
+const argOut: string = argv._[1] || argv['output']
+
+function readInput(): Promise<string> {
```
bcherny: Can make this `async`

README.md
erykpiast: :+1:

src/cli.ts
erykpiast: I think it is possible to get error when writing to stdout, isn't it?

src/cli.ts
erykpiast: never, it doesn't make sense :D

src/cli.ts
erykpiast: I'm not sure it does make sense. The code could be simplified to
```
function readInput(): Promise<string> {
  if (!argIn) {
    return new Promise(stdin);
  }

  return readFile(argIn)
}
```
Is it the case for `async`?

bcherny:
Looking good- thanks @erykpiast!

PR #65, opened 2017-03-07
Re-export for interfaces

monolithed:


bcherny:
Thanks for the contributions @monolithed!

PR #89, opened 2017-05-16
remove duplicated block of logic

vatula:


bcherny:
Nice catch - thanks for the contribution @vatula!

PR #99, opened 2017-06-21
Make keys in Options *actually* optional

grantila:


src/index.ts
```diff
@@ -13,15 +13,15 @@ import { validate } from './validator'
 export { EnumJSONSchema, JSONSchema, NamedEnumJSONSchema } from './types/JSONSchema'
 
 export interface Options {
-  bannerComment: string,
```
bcherny: Hi @grantila - sorry for the delay, I'm travelling. Instead of this, would you mind just adding `Partial<Options>` on these lines instead?

- https://github.com/bcherny/json-schema-to-typescript/blob/98e1127bc0a6361f10dce73e757191eb668c407d/src/index.ts#L45
- https://github.com/bcherny/json-schema-to-typescript/blob/98e1127bc0a6361f10dce73e757191eb668c407d/src/index.ts#L65

grantila:
Closing in favor of #102

Issue #16, opened 2016-09-03
Batch mode for generating interfaces for several JSON-schemas at once


bcherny:
To avoid generating the same interface twice.

https://github.com/bcherny/json-schema-to-typescript/pull/15#discussion_r77439790

erykpiast:
So how the API may look like? `json2ts -i *.json -o ???`. Should we allow to specify some kind of renaming pattern (like `sed` syntax) or just concatenate all interfaces to one `ts` file when output file is specified and create `[orig_name_wo_extension].d.ts` files in case it isn't?

bcherny:
Glob syntax sounds good to me. 

I would separate the work into 2 parts:

- Part 1 (easy) - each input file maps to an output file (user can take care of concating), no optimization (TS takes care of type merging, so duplicates are fine)
- Part 2 (harder) - each input file maps to an output file, common types are pulled out into a common file to avoid duplication

erykpiast:
I like the split :+1:.

So, for the first step, results would be like 
```
$ ls *.json
a.json b.json c.json
$ json2ts *.json
$ ls *.d.ts
a.d.ts b.d.ts c.d.ts
``` 
no matter if `a` `b` or `c` contain different or equal interfaces, right?

How do we want to prevent people from specifying output file for multiple inputs? Just throw an error?

bcherny:
Exactly. Error or warning is probably fine. I agree error might be better to keep the API as strong as possible.

tomekp:
I expect a little different behavior:
I want to be able to generate multiple `*.ts` files, one for each defined interface (from each schema). So if some type would be generated many times it would just overwrite the same file.

It seems I'd need to add a function to generator that returns a dictionary like `{"file name for type": "file content"}` (+ change all `declareXxx` functions to return an array of strings instead of returning concatenated string, and figure out imports).

The tricky part seems to be generating file names (for `SomeType` in schema I want to get `some-type.model.ts` file) so I'd expect user to provide file name generating function.

If that's OK for you I'd start working on a PR providing that.

bcherny:
@tomekp I'm not convinced that approach is better. Ostensibly if someone has already taken the trouble to put, say, definitions A B and C in schema X and D E and F in schema Y, for their case this is a natural way to modularize and group their schemas. So when we generate TypeScript code from those schemas, it seems natural to put schemas A B and C in X.d.ts and D E and F in Y.d.ts.

[...]

PR #101, opened 2017-06-22
compileFromDir (aka batch compile)

tomholford:
Consider this a WIP implementation to address #16 

Call `compileFromDir` with a list of filenames and the optional options hash (same method signature as `compileFromFile`)

todo:
- [ ] Tests 
- [ ] CLI support


tomholford:
@bcherny Both :) Sketched out an idea to address #16 and wanted to see if it's congruent with your roadmap before investing more time into it. Further, wanted to solicit feedback on the interface - i.e. should this be available via CLI? And if so, what would the command look like?

For example:
```
json2ts -d schemas/models/ -R
```
where `-d` indicates the directory for compilation and `-R` indicates to traverse the directory recursively

bcherny:
@tomholford See this comment and the one after it - does that seem reasonable to you?

https://github.com/bcherny/json-schema-to-typescript/issues/16#issuecomment-282951441

bcherny:
@tomholford Closing for inactivity. Feel free to reopen when/if you get back to this.

tomholford:
@bcherny Thanks, will re-open later

Issue #104, opened 2017-07-13
Typenames of NAMED_ENUM schemas should be derived from title when present


aparna-sath:
```json
"gender": {
  "type": "string",
  "title": "GenderType",
  "tsEnumNames": ["MALE", "FEMALE"],
  "enum": [0, 1]
}
```
This should generate a typescript enum called GenderType, rather than Gender.

bcherny:
PRs are welcome!

bcherny:
Fixed in f9c8624.

```sh
+ json-schema-to-typescript@4.5.0
```

PR #105, opened 2017-07-13
Use title as the typename for a named enum, when available

aparna-sath:
https://github.com/bcherny/json-schema-to-typescript/issues/104


bcherny:
Thanks for the contribution @aparna-sath!

PR #106, opened 2017-07-24
Fix string literal followed by array

jefbarn:
Changes this,
```blah: "abc"[]```
to this
```blah: ("abc")[]```
to fix a compiler error.

Also added helper library so that install and test work on Windows.

### CODE REVIEW

**Comment by bcherny** [comment]
This looks great - thanks for the contribution @jefbarn!

**Comment by bcherny** [comment]
```sh
+ json-schema-to-typescript@4.5.2
```

PR #112, opened 2017-08-15
Basic support for patternProperties

dustinblackman:
## Overview

This is a quick fork I put together to add basic for `patternProperties` by adding `[k: string]` as the key. While this doesn't do any of the magic `patternProperties` should do, it at least gives typing support for an example such as this instead of interfaces not being filled at all:

```json
{
  "id": "example.json",
  "type": "object",
  "patternProperties": {
    ".+": {
      "properties": {
        "reference": {
          "type": "string"
        },
      },
      "required": [
        "reference"
      ]
    }
  }
}
```

```typescript
export interface Example {
  [k: string]?: {
    reference: string;
  };
}
```

I'm assuming people who are moving over to using Typescript interfaces are likely still using JSONSchema somewhere else in the stack (such as tests) so patternProperties is still used correctly that way. Possibly having a section called something like `minimum support` in the README could be useful. If something like this doesn't fit the path of the project however, feel free to close! Just wanted to share incase it could be of use.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -252,11 +252,22 @@ function parseSchema(
   processed: Processed
 ): TInterfaceParam[] {
 
-  const asts = map(schema.properties, (value, key: string) => ({
+  let asts = map(schema.properties, (value, key: string) => ({
     ast: parse(value, rootSchema, key, true, processed),
     isRequired: includes(schema.required || [], key),
     keyName: key
   }))
+
+  // Very basic support for patternProperties.
+  if (schema.patternProperties) {
+    const patternProps = map(schema.patternProperties, (value, key: string) => ({
+      ast: parse(value, rootSchema, key, true, processed),
+      isRequired: includes(schema.required || [], key),
+      keyName: '[k: string]'
+    }))
+
+    asts = asts.concat(patternProps)
+  }
 
   // handle additionalProperties
   switch (schema.additionalProperties) {


bcherny:
@dustinblackman I'm not sure this is the correct usage for `patternProperties`. The regex `.+` happens to translate to `[k: string]`, but this is not the case for the vast majority of possible regexes. Not only that, but `.+` is a pathological case: if someone is using that particular regex, they're better off just using `additionalProperties`, which would communicate the intent more clearly and is supported by JSTT already.

That said, if you do want to support that particular regex, you would have to check that the regex is equivalent to "everything". This is not trivial to verify, as there are an infinite number of regexes that satisfy that equivalence.

With your approach, if I have the `patternProperties` key `\d+`, JSTT would generate a schema that would allow the key `"abc"`, which is not valid. For this reason, I made the call to not support `patternProperties` at all - it's better to always generate valid interfaces than to sometimes generate invalid ones.

What do you think?

---------

From the [spec](https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.3):

> **5.3.  patternProperties**
>   This attribute is an object that defines the schema for a set of
   property names of an object instance.  The name of each property of
   this attribute's object is a regular expression pattern in the ECMA
   262/Perl 5 format, while the value is a schema.  If the pattern
   matches the name of a property on the instance object, the value of
   the instance's property MUST be valid against the pattern name's
   schema value.

adamduren:
@dustinblackman is this supported within this library?

adamduren:
I'm essentially looking for a way to express that an object can contain any arbitrary key as long as it's a certain type. This is a common pattern in Firebase and I'm looking for a way to represent Db schema. Currently objects have any type when they are generated and I'm looking to generator something like.

```
export interface IDb {
  users: {
    [k: string]: IUserProfile;
  };
}

```

bcherny:
@adamduren You're looking for `additionalProperties`. See example usage [here](https://github.com/bcherny/json-schema-to-typescript/blob/master/test/e2e/additionalProperties.ts).

adamduren:
Thanks! I missed that earlier. Thought it could only be a boolean.

PR #114, opened 2017-09-26
Fix cwd handling

JochenDiekenbrock:
I found a bug in compileFromFile today. The cwd property in the options is overridden by the directory of the base schema file.
To reproduce the bug, I added the test test/testCompileFromFile.ts. This test reads the file test/resources/other/ReferencingType.json containing "$ref": "ReferencedType.json"
By providing the option {cwd: './test/resources'} in the call to compileFromFile the reference should be resolved correctly.

Without my fix, the test fails because json-schema-to-typescript tries to read test/resources/other/ReferencedType.json

With the fix, the test is ok. Please integrate the fix into the next version. Thanks!


bcherny:
This looks great - thanks for the contribution @JochenDiekenbrock!

bcherny:
Published `+ json-schema-to-typescript@4.6.3`

JochenDiekenbrock:
Thank you Boris, for merging the fi and for json-schema-to-typescript in general!

Issue #129, opened 2017-11-11
Generating patternProperties Objects


dekryptic:
Is this schema:
```
{
  "id": "Parent",
  "type": "object",
  "patternProperties": {
    "^[a-zA-Z]+": {
      "id": "Child",
      "type": "object",
      "properties": {
        "aProperty": { "type":"string" }
      }
    }
  }
}
```
not supposed to generate a Child interface? I tested it out on https://bcherny.github.io/json-schema-to-typescript-browser/ and it only returned a Parent interface, i.e.
```
export interface Parent {
  [k: string]: any;
}
```

If not, is there any way to force json-schema-to-typescript to generate an interface for ALL objects, regardless of whether or not a specific portion of the JSON Schema specification can be supported in Typescript?

bcherny:
Hi @dekryptic! What exactly would you expect the output to be in this case?

dekryptic:
Would something like this:
```
export interface Parent {
  [k: string]: Child;
}
export interface Child {
  aProperty?: string;
  [k: string]: any;
}
```
be reasonable?

bcherny:
I see. I believe this is a duplicate of https://github.com/bcherny/json-schema-to-typescript/issues/124. Feel free to reopen if you feel otherwise.

Issue #124, opened 2017-10-23
Inspect Pattern Properties just for recursive definitions.


robertmassaioli:
I just want to make it initially clear: I'm not looking for you to generate code for patternProperties. 

However, pattern properties do reference definitions that would be good to generate samples for. For example, in swagger we have the following definition:

      "paths": {
        "type": "object",
        "description": "Relative paths to the individual endpoints. They must be relative to the 'basePath'.",
        "patternProperties": {
          "^x-": {
            "$ref": "#/definitions/vendorExtension"
          },
          "^/": {
            "$ref": "#/definitions/pathItem"
          }
        },

Since this definition relies exclusively on patternProperties the generated object looks like this:

    /**
     * Relative paths to the individual endpoints. They must be relative to the 'basePath'.
     */
    export interface Paths {
    
    }

I think that we can all agree that this is not very useful. Frankly, a more useful implementation would be:

    /**
     * Relative paths to the individual endpoints. They must be relative to the 'basePath'.
     */
    export interface Paths {
        [name: string]: VendorExtension | PathItem;    
    }

But, like I said, I don't really want to solve that problem here. **The real problem is that VendorExtension and PathItem are never generated at all.**

This means that I can't even do some minor edits, after the fact, to the generated code, to make it behave as I expect. It would be really could if you could scan the patternProperties just to continue the code generation so the resultant file was easier to modify to be correct. 

Would that be an acceptable compromise? That way, you don't need to solve the patternProperties problem but users of your excellent tool could get 95% of the way to fully generated code instead of 60% of the way there in examples that make heavier use of patternProperties.

bcherny:
Hi @robertmassaioli!

```ts
/**
 * Relative paths to the individual endpoints. They must be relative to the 'basePath'.
 */
export interface Paths {
    [name: string]: VendorExtension | PathItem;    
}
```

For the record, the problem with this interface is it permits assigning a `VendorExtension` to a key that matches `^/`, and a `PathItem` to a key that matches `^x-` (both of which are invalid according to the JSON-Schema). It would also allow assigning either a `VendorExtension` or a `PathItem` to a key that matches neither `^/` nor `^x-` (eg. the key `foo`), which is also invalid.

That said, generating interfaces for `VendorExtension` and `PathItem` does seem like a good idea. The downside is it's unclear where/how they are used, and the upside is - as you said - we should generate as much as possible while remaining correct.

For Phase 1, let's generate interfaces for `patternProperties`.

[...]

PR #138, opened 2018-01-11
Add partial support for patternProperties

StefanoDalpiaz:
With this PR I am proposing an implementation that adds support for `patternProperties`. 

While the Typescript notation cannot fully express the constraints that this feature provides, I am trying to make the most of the information that can be handled by Typescript.

The idea is to accept the fact that we cannot translate the pattern matching, but we can still translate the hierarchical information between types.

For example let's consider the following schema:

```json
{
  "$schema": "http://json-schema.org/draft-04/schema",
  "type": "object",
  "patternProperties": {
    "^a.*": {
      "type": "string"
    },
    "^b.*": {
      "type": "number"
    }
  },
  "additionalProperties": false
}
```

In the above example, while we cannot express the pattern matching (i.e. whether the key starts with "a" or "b"), we can still provide the information that the values can be either a string or a number:

```typescript
export interface Foo {
  [k: string]: string | number;
}
```

In a more complex scenario, the implementation also considers:

- the existence of other "normal" properties (the indexed values types should include all the possible types of the normal properties, according to the Typescript specs)
- whether those properties are optional or not (if they are, the indexed values should also include `undefined`, for the same reason as above)
- whether additional properties are allowed or not (if they are, the type `any` should be added to the list of types for the indexed values - this would actually make the rest of the definitions irrelevant from a type-checking point of view, but I still think it's useful for humans to see the possible types)

Here is an example that includes all the conditions above (I added more permutations as test files as part of this PR):

```json
{
  "$schema": "http://json-schema.org/draft-04/schema",
  "type": "object",
  "patternProperties": {
    "^a.*": {
      "type": "string"
    },
    "^b.*": {
      "type": "number"
    }
  },
  "properties": {
    "something": {
      "type": "array",
      "items": {
        "type": "string"
      }
    }
  },
  "additionalProperties": false
}
```

```typescript
export interface Foo {
  [k: string]: string | number | string[] | undefined | any;
  something?: string[];
}
```


bcherny:
@StefanoDalpiaz Thanks for the contribution! This has been covered [several](https://github.com/bcherny/json-schema-to-typescript/issues/129) [times](https://github.com/bcherny/json-schema-to-typescript/issues/124) in issue already, and I believe it is beyond the scope of what JSON-Schema-to-TypeScript should do because it will produce typings that permit invalid JSON-Schemas.

See discussion [here](https://github.com/bcherny/json-schema-to-typescript/issues/124#issuecomment-338731590) - what do you think?

bcherny:
@StefanoDalpiaz Closing for inactivity. Please reopen if you'd like to continue the discussion.

PR #144, opened 2018-02-08
Implement partial support for patternProperties

qix:
Hello!

Sorry about the code style, I tried to keep it the same but it's quite different from anything I've seen before. Happy to change anything as needed.

As for the change: in real world use cases `patternProperties` is often used to enforce some regular expression on keys (as evidenced by `heroku`/`jsonStat` real world test cases).

In those cases it's **super** useful to atleast have partial patternProperties support. This change implements patternProperties under a rather strict condition:

- additionalProperties is not set (or is false)
- there is only a single patternProperty defined

This improves some real world use cases while still producing correct output.

Note: There are two commits here, I had to first update ava snapshots in order to produce a clean commit for review.


bcherny:
Thanks for the contribution @qix! However, I believe this will introduce incorrect behavior, and I'm not convinced that that behavior is worth having this feature. See full discussion here: https://github.com/bcherny/json-schema-to-typescript/pull/112#issuecomment-322600537.

What do you think?

qix:
I'm not super convinced. In the case that you want all keys to map to a type of value `additionalProperties: ...` works, but what about only certain keys?

The commit has some good examples of this:
https://github.com/bcherny/json-schema-to-typescript/pull/144/commits/a68c44764a7a42cdad8597de3b654d9800382839#diff-7b1a5e1c6a45babf7dcc4aaf9c57995dL2121

That's just one, but in that example environment variables are capital letters only in the keys, and the value must be strings. After the change the type is reduced from `[key: string]: any` to `[key: string]: string`. While it allows you to create invalid objects, the types are atleast a little stricter.

It's even more important when you consider more complex objects. See [again from one of your real world test cases]:
https://github.com/bcherny/json-schema-to-typescript/pull/144/commits/a68c44764a7a42cdad8597de3b654d9800382839#diff-7b1a5e1c6a45babf7dcc4aaf9c57995dL4378

Just because the JSONSchema used a patternProperies to validate key names, before this change it essentially mapped the entire type to `any`. Post change you get full types on the values, even if you can supply invalid keys.

bcherny:
@qix I think you might be right. Given there's been so much demand for this, let's go for it. Will do a more thorough code review this weekend :)

src/parser.ts
```diff
@@ -284,18 +284,27 @@ function parseSchema(
...
     asts = asts.concat(map(schema.patternProperties, (value, key: string) => {
       let ast = parse(value, options, rootSchema, key, true, processed, usedNames)
       let comment = `This interface was referenced by \`${parentSchemaName}\`'s JSON-Schema definition
 via the \`patternProperty\` "${key}".`
       ast.comment = ast.comment ? `${ast.comment}\n\n${comment}` : comment
       return ({
         ast,
-        isPatternProperty: true,
-        isRequired: includes(schema.required || [], key),
+        isPatternProperty: singlePatternProperty ? false : true,
```
bcherny: Use shorthand (same thing below)

```diff
- isPatternProperty: singlePatternProperty ? false : true,
+ isPatternProperty: !singlePatternProperty,
```

test/__snapshots__/test/test.js.md
```diff
@@ -850,6 +889,197 @@ Generated by [AVA](https://ava.li).
...
+     * via the `patternProperty` "^[0-9]+".␊
+     */␊
+    export type NumberChild = number;␊
+    /**␊
+     * This interface was referenced by `Parent`'s JSON-Schema definition␊
+     * via the `patternProperty` "^[a-zA-Z]+".␊
+     */␊
+    export type StringChild = string;␊
+    ␊
+    export interface Parent {}␊
```
bcherny: Is this the right output? I would expect

```ts
export interface Parent {
  [k: string]: NumberChild | StringChild
}
```

bcherny:
@qix This looks great overall, and is a well thought out way of doing this!

Please also add tests for:

1. A parent schema that has both `additionalProperties=true` and `patternProperties` set
2. A parent schema that has both `additionalProperties` set to a schema and `patternProperties` set

bcherny:
@qix Just checking in - any updates on this?

test/__snapshots__/test/test.js.md
qix: That would be better, but I didn't manage to get that working. For now the set of rules are quite conservative.

qix:
@bcherny I've fixed the requested issue and rebased on top of latest master

test/__snapshots__/test/test.js.md
```diff
@@ -3949,7 +4047,13 @@ Generated by [AVA](https://ava.li).
       /**␊
        * hash mapping process type names to their respective command␊
        */␊
-      process_types?: {};␊
+      process_types?: {␊
+        /**␊
+         * This interface was referenced by `undefined`'s JSON-Schema definition␊
```
bcherny: Why is this `undefined`?

test/__snapshots__/test/test.js.md
bcherny: Tracking this in https://github.com/bcherny/json-schema-to-typescript/issues/160

test/__snapshots__/test/test.js.md
bcherny: Tracking in https://github.com/bcherny/json-schema-to-typescript/issues/161

bcherny:
Looking good. Thanks for the contribution @qix!

bcherny:
Published 5.5.0.

PR #147, opened 2018-02-15
add declarations for enums referenced by arrays (fix #146)

### LINKED ISSUE #146: Enums $refd from items are not included in output
Version: 5.3.3

Example:

```json
{
  "$schema": "http://json-schema.org/draft-04/schema",
  "title": "MyInterface",
  "type": "array",
  "items": { "$ref": "#/definitions/MyEnum" },
  "definitions": {
    "MyEnum": {
      "title": "MyEnum",
      "enum": ["VALUE1", "VALUE2"],
      "tsEnumNames": ["Value1", "Value2"]
    }
  }
}
```

Output:
```typescript
/**
 */

export type MyInterface = MyEnum[];
```

victorandree:
This makes sure that array and tuple `items` are searched for standalone enum definitions to generate. Before, the array would used the enum's type name, but no enum definition would be generated (See #146).

While writing this, I found a related bug: If you don't provide an explicit title for an enum defined in `items`, the type will have no name so the generator crashes (in `generateType` because `let type = generateType(ast.params, options)` is `undefined`). It's not obvious what an autogenerated enum name should be (it can't be inferred from the property name). Example:

```json
{
  "type": "array",
  "items": {
    "enum": [1, 2, 3],
    "tsEnumNames": ["One", "Two", "Three"]
  }
}
```

Ideas:

1. Either validate that an enum so defined has a title.
2. Infer a name based on the array, for example `MyArrayEnum` or `MyArrayItem`. I don't think this is very useful, however.


bcherny:
Published `json-schema-to-typescript@5.4.0`.

PR #148, opened 2018-03-13
Prevent logging from clashing with `debug` module.

jrylan:
The current implementation clashes with https://www.npmjs.com/package/debug causing unexpected terminal output when a user may have been specifying a pattern intended for the `debug` module.

`debug` has nearly 75 million download per month, so quite lot of people could hit this.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -77,7 +77,7 @@ export function error(...messages: any[]) {
 }
 
 export function log(...messages: any[]) {
-  if (process.env.DEBUG) {
+  if (process.env.VERBOSE) {
     console.info(whiteBright.bgCyan('debug'), ...messages)
   }
 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -76,9 +76,8 @@ export function error(...messages: any[]) {
   console.error(whiteBright.bgRedBright('error'), ...messages)
 }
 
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -76,8 +76,9 @@ export function error(...messages: any[]) {
   console.error(whiteBright.bgRedBright('error'), ...messages)
 }
 
+const debugEnabled = process.env.DEBUG !== undefined && process.env.DEBUG.includes('json2ts')
```
jrylan: I used `.includes` here instead of a direct equality check in order to make this compatible with the `debug` module, since the `DEBUG` env var as used in that package encourages users to separate namespaces or patterns using a comma or space, so in my case I might do something like the following:

env DEBUG=react,request,json2ts

bcherny:
Thanks for the heads up @jrylan! Can you just rename `DEBUG` to `VERBOSE` instead?

jrylan:
@bcherny Would you like everything as it was originally, but just with `VERBOSE` instead of `DEBUG`? 

So in order for someone to debug `json-schema-to-typescript` they would be writing using `env VERBOSE=true` like you previous had for `env DEBUG=true`?

bcherny:
Exactly. Does that work for you?

jrylan:
@bcherny sounds good to me. Thank you for the great module!

bcherny:
Published `json-schema-to-typescript@5.3.4`

PR #168, opened 2018-07-11
add `tsType` schema extension to allow custom types

sokra:
This add a feature:

`tsType` can be used in a schema (similar to `tsEnumNames`) to specify the generated type. This is useful when using custom schema extensions which are not supported by `json-schema-to-typescript`. It can also be used in cases where the json-schema is not expressive enough.

Example schema:

``` json
{
  "instanceOf": "Function",
  "tsType": "(a: number, b: number) => number"
}
```

Parameter is passed into the generated code without modification.


bcherny:
@sokra Thanks for the contribution. This is clean, and could be a nice escape hatch for people that use non-standard JSON-Schema extensions.

That said, what is the specific use case you have in mind for this? If you're typing TypeScript functions, why use JSON-Schema at all, instead of using a `.d.ts` or `.ts` file as your source of truth?

gunnar-t:
I would love to use this feature!

My use case is indead ONE source of truth: the nodejs backend defining the basic api structure via json schemas. These schemas are then used for generating the typescript interfaces for the frontend. So I only have one place to configure data structures. But I need to add the "any" type in the d.ts file for the frontend....

Nice work by the way.

bcherny:
Hey @sokra, I'm just getting around to this. Do you mind rebasing, then I can merge. Or, turn on push access for reviewers.

sokra:
> That said, what is the specific use case you have in mind for this? If you're typing TypeScript functions, why use JSON-Schema at all, instead of using a `.d.ts` or `.ts` file as your source of truth?

Schemas are more expressive and can be used for runtime validation. But schema lacks validation for functions and types. This additional would allow to add the missing information to the schema.



> Do you mind rebasing, then I can merge. Or, turn on push access for reviewers.

I'll rebase.

sokra:
rebased

sokra:
#191 fixes a bug with this PR

PR #178, opened 2018-08-22
Refactor typeOfSchema.ts

qm3ster:
It benchmarks better, and it is clearer what takes priority over what.
A further performance optimization could be making a left and a right array from `propertyMap` (manually or at module parse time), so instead of
```ts
const firstPropertyMatched = propertyMap.find(([key]) =>
    schema.hasOwnProperty(key)
  )
```
we end up doing
```ts
const firstPropertyIndex = propertyKeys.findIndex(key => schema.hasOwnProperty(key) )
const val = firstPropertyIndex !== -1
    ? propertyValues[firstPropertyIndex]
    : {/* other stuff */}
```

There is currently one difference in undocumented behavior of the function exposed by tests:

When **`schema`** argument passed is literally `[ 'firstName', 'lastName' ]`, the old function returned `'ANY'` while the new returns `'OBJECT'`.
This doesn't actually fail any tests, and might be an error in the test.
I didn't handle this in any special way because it is an incorrect invocation of the function.

TypeScript 2.8 is required to type this code (due to Conditional Types in `KnownKeys`), but that doesn't affect output or anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "shx": "^0.2.2",
     "tsify": "^3.0.3",
     "tslint": "^5.8.0",
-    "typescript": "^2.6.1"
+    "typescript": "^2.8.0"
   },
   "ava": {
     "files": [
--- a/src/typeOfSchema.ts
+++ b/src/typeOfSchema.ts
@@ -5,6 +5,16 @@ import { JSONSchema, SCHEMA_TYPE } from './types/JSONSchema'
  * Duck types a JSONSchema schema or property to determine which kind of AST node to parse it into.
  */
 export function typeOfSchema(schema: JSONSchema): SCHEMA_TYPE {
+  const oldJudgement = typeOfSchemaOld(schema)
+  const newJudgement = typeOfSchemaNew(schema)
+  if (newJudgement !== oldJudgement) {
+    console.warn(newJudgement, oldJudgement, typeof schema)
+    console.log(schema)
+  }
+  return oldJudgement
+}
+
+function typeOfSchemaOld(schema: JSONSchema): SCHEMA_TYPE {
   if (schema.allOf) return 'ALL_OF'
   if (schema.anyOf) return 'ANY_OF'
   if (schema.oneOf) return 'ONE_OF'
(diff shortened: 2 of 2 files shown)


qm3ster:
@bcherny please have a look if you have the time

bcherny:
Hey @qm3ster, I'm just getting a chance to look at this. First of all, thanks for taking the time to make this pull request.

You mentioned this version "benchmarks better" - would you be able to share some of those benchmarks? Your version is quite a bit more complex than the previous one, and I'd love to better understand the perf gains we'd get with it.

qm3ster:
@bcherny Oh, sorry.
This isn't meant to be merged, this runs both the old and the new version side by side.
The reason it looks more complex is I didn't remove the old version.
I just wanted to know if you are okay with this style, in which case I'll fix up the PR.
If you check this branch out as it is now, you can verify for yourself, for example by running the tests, or using this on your own codebase, that the output matches.

Architecturally, the change is the following:
The old version looked a homogenous block of early-return.
But in reality, there was:
### 1. Checking for property existence section:
```ts
  if (schema.allOf) return 'ALL_OF'
  if (schema.anyOf) return 'ANY_OF'
  if (schema.oneOf) return 'ONE_OF'
  if (schema.items) return 'TYPED_ARRAY'
  // the following two are logically related, not just ordered by priority
  if (schema.enum && schema.tsEnumNames) return 'NAMED_ENUM'
  if (schema.enum) return 'UNNAMED_ENUM'
  if (schema.$ref) return 'REFERENCE'
```
Which I replaced with this table:
```ts
const propertyMap: [KnownKeys<JSONSchema>, SCHEMA_TYPE | SchemaIdentifier][] = [
  ['allOf', 'ALL_OF'],
  ['anyOf', 'ANY_OF'],
  ['oneOf', 'ONE_OF'],
  ['items', 'TYPED_ARRAY'],
  [
    'enum',
    schema =>
      schema.hasOwnProperty('tsEnumNames') ? 'NAMED_ENUM' : 'UNNAMED_ENUM'
  ],
  ['$ref', 'REFERENCE']
]
```
which is used by this simple find:
```ts
  const firstPropertyMatched = propertyMap.find(([key]) =>
    schema.hasOwnProperty(key)
  )
```
This can be further optimized by unzipping the tuples into two arrays at module init time, as described in the PR body.
### 2. Switch on `typeof`, which has one nested condition that allows function to continue:
```ts
  switch (schema.type) {
    case 'string': return 'STRING'
    case 'number': return 'NUMBER'
    case 'integer': return 'NUMBER'
    case 'boolean': return 'BOOLEAN'
    case 'object':
      if (!schema.properties && !isPlainObject(schema)) {
        return 'OBJECT'
      }
      break
    case 'array': return 'UNTYPED_ARRAY'
    case 'null': return 'NULL'
    case 'any': return 'ANY'
  }
```
I replaced this block with this dictionary:
```ts
const typeLookup: Record<
  Strings<JSONSchema['type']>,
  SCHEMA_TYPE | SchemaIdentifier
> = {
  string: 'STRING',
  number: 'NUMBER',
  integer: 'NUMBER',
  boolean: 'BOOLEAN',
  object: anyMoreProps,
  array: 'UNTYPED_ARRAY',
  null: 'NULL',
  any: 'ANY'
}
```
Property lookups are faster than basic `switch`, and this again converts it from control flow to a data structure, which is easier to reason about. There is also no nested case of breaking out of the lookup, the logic flow is firmly contained in the `object` case.

### 3. `typeof schema.default`
Same as above.
```ts
  switch (typeof schema.default) {
    case 'boolean': return 'BOOLEAN'
    case 'number': return 'NUMBER'
    case 'string': return 'STRING'
  }
```
```ts
const defaultMap: Record<string, SCHEMA_TYPE | undefined> = {
  boolean: 'BOOLEAN',
  number: 'NUMBER',
  string: 'STRING'
}
```

## Performance
The benchmark was rather contrived: timing the sum of all invocations of the old version vs the new version with them running side by side on the test suite.
I suppose most if not all the performance improvement comes from the lookups instantiated outside the function. Performance was not really a concern to begin with, but on the biggest schemas it can be .1ms spent in this function instead of .2ms :pizza:

bcherny:
@qm3ster Thanks for looking into this, and for the thorough response. I think the existing `switch is significantly easier to understand than the new solution, so let's keep this as-is.

Issue #179, opened 2018-08-24
Allow passing options to $RefResolver


qm3ster:
I need to pass some [*Custom Resolvers*](https://github.com/BigstickCarpet/json-schema-ref-parser/blob/master/docs/plugins/resolvers.md#custom-resolvers) to the resolver, to handle loading schemas from non-json files.

PR #180, opened 2018-08-24
Pass options to $ref-resolver

qm3ster:
fixes https://github.com/bcherny/json-schema-to-typescript/issues/179


qm3ster:
Also, why are you instantiating a resolver?
https://github.com/bcherny/json-schema-to-typescript/blob/1959154c9e11fd3e67d027d995520f35a79b32a2/src/resolver.ts#L8-L9
It never leaves the function, so it is just disposed of, the extra properties on the instance are never accessible.
Maybe should use static method instead, or make one instance on module init and reuse it?

qm3ster:
Now that I familiarized myself with the internals of `$RefParser`, that's actually what the static method is doing - making a new instance, the method on it, and disposing of it.

qm3ster:
Will do!

qm3ster:
Here it is on top of master, with passing tests, with minimal change to the typedef to expose the `Options` type I need to check.

PR #186, opened 2018-09-17
add .editorconfig

iwan-aucamp-cs:
[editorconfig](https://editorconfig.org/) for your project to make it easier for people to ensure they use right style.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,4 +8,4 @@ tab_width = 2
 end_of_line = lf
 charset = utf-8
 insert_final_newline = true
-trim_trailing_whitespace = false
+trim_trailing_whitespace = true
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+tab_width = 2
+end_of_line = lf
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = false
```
bcherny: Should this be `true`?

iwan-aucamp-cs: Indeed it should - changed it.

iwan-aucamp-cs:
I have added `.editorconfig` and to `.npmignore` and changed `insert_final_newline` to `true`

PR #187, opened 2018-09-17
Make typescript strict

iwan-aucamp-cs:
Change `tsconfig.json` to use strict mode for compiling typescript and fix the only error which presents when this is used.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,13 +7,11 @@
     "module": "commonjs",
     "moduleResolution": "node",
     "noFallthroughCasesInSwitch": true,
-    "noImplicitAny": true,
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -14,10 +14,11 @@
     "preserveConstEnums": true,
     "sourceMap": true,
     "strictNullChecks": true,
```
bcherny: This change means you can rm lines 10 and 16

iwan-aucamp-cs: I removed the redundant `strictNullChecks` and `noImplicitAny` options from `tsconfig.json`

PR #191, opened 2018-09-20
fix bug when optimizing custom types

sokra:
rename param to params for CUSTOM_TYPE

Otherwise the optimizer doesn't work correctly and merges all custom types into one

Note: The optimizer always stringifies `params`, so it has to be this name. I wasn't aware of this when writing the problematic code.


bcherny:
Published `6.0.1`

sokra:
Thanks

> Published `6.0.1`

I think you forgot to git push the commit and tag.

Issue #181, opened 2018-08-28
Add option to control type naming.


iwan-aucamp-cs:
Currently type names are derived as follow: 

```
/**
 * Compute a schema name using a series of fallbacks
 */
function standaloneName(
  schema: JSONSchema,
  keyNameFromDefinition: string | undefined,
  usedNames: UsedNames
) {
  let name = schema.title || schema.id || keyNameFromDefinition
  if (name) {
    return generateName(name, usedNames)
  }
}
```

This is not ideal if title is used for other purposes - I would like to rather specify the priority - so for example `--typeNames id,title,definition` should use either id, and if id is not set use title and then finally fall back to definition.

bcherny:
Hey there! This is a pretty unusual behavior to customize. Could you give more context for why you want a different name resolution algo for your use case?

iwan-aucamp-cs:
We are using `https://mozilla-services.github.io/react-jsonschema-form/` to generate forms from JSON schema - and this is using title in the output form. This seems to be what title was intended for:

[JSON Schema: interactive and non interactive validation draft-fge-json-schema-validation-00](https://tools.ietf.org/html/draft-fge-json-schema-validation-00)

> 
> 6.1.  "title" and "description"
> 
> 6.1.1.  Valid values
> 
>    The value of both of these keywords MUST be a string.
> 
> 6.1.2.  Purpose
> 
>    Both of these keywords can be used to decorate a user interface with
>    information about the data produced by this user interface.  A title
>    will preferrably be short, whereas a description will provide
>    explanation about the purpose of the instance described by this
>    schema.
> 
>    Both of these keywords MAY be used in root schemas, and in any
>    subschemas.

Using something which is likely to contain human readable text for type name generation is not ideal - but I would much rather prefer for my case to always use `keyNameFromDefinition` as this will provide overall better consistency from code point of view - if someone adds or changes ID/Title then the type names will remain consistent and it simplifies the heuristics for determining what the name of a type will be.

The reason why this should be configurable is because changing it outright to use `id` or `keyNameFromDefinition` instead of `title` will likely break things for people relying on current behavior.

bcherny:
@iwan-aucamp-cs That sounds fair to me! Feel free to submit a PR introducing a new option for this, `customName`, which should be a function that takes `(id, title, keyNameFromDefinition)` and returns a string. If that option is passed in, we can use it when computing the name in the  `let name` line.

ankon:
[...]

PR #215, opened 2019-01-13
Add customName option and namePriority for CLI

aucampia:
The `customName` option takes takes a function that returns the name to assign to schema object and `namePriority` command line option that makes use of this mechanism. Related to #181


src/cli.ts
```diff
@@ -81,6 +96,11 @@ Boolean values can be set to false using the 'no-' prefix.
       Prettier configuration
   --unreachableDefinitions
       Generates code for definitions that aren't referenced by the schema
+  --namePriority={${values(CliNameSource).join(',')}},...
+      A commma seperated list of priorties for type naming.
```
ankon: Typo/Nit-picking: "comma-separated", "priorities"

src/parser.ts
```diff
@@ -68,44 +68,45 @@ function parseNonLiteral(
 ) {
 
   log(whiteBright.bgBlue('parser'), schema, '<-' + typeOfSchema(schema), processed.has(schema) ? '(FROM CACHE)' : '')
+  const customName = () => { return ( options.customName ? options.customName({schema, keyName, definitionKeyName: keyNameFromDefinition, usedNames}) : undefined ) }
```
ankon: I think for consistency it would be nice to pull up the `customName() || standaloneName()` parts into a `name` here as well, in the same way it is done later?

src/cli.ts
aucampia: thanks, I fixed it.

src/parser.ts
aucampia: customName and standaloneName potentially change state (i.e. it marks a name as being used) and the result would not be used in cases where `newInterface` is called. I guess though I could change `newInterface` signature and just pass the generated name through, but there is still potential for problems. The other option is to make a function wrapper `let name = () => { return customName() || standaloneName(schema, keyNameFromDefinition, usedNames)! }` - will look into this maybe a bit later.

aucampia: @ankon I have simplified it slightly - have a look and tell me what you think.

src/cli.ts
```diff
@@ -81,6 +96,11 @@ Boolean values can be set to false using the 'no-' prefix.
       Prettier configuration
   --unreachableDefinitions
       Generates code for definitions that aren't referenced by the schema
+  --namePriority={${values(CliNameSource).join(',')}},...
```
bcherny: Could we keep this PR to `customName`, and avoid introducing a new `namePriority` option?

src/parser.ts
```diff
@@ -68,44 +88,44 @@ function parseNonLiteral(
 ) {
 
   log(whiteBright.bgBlue('parser'), schema, '<-' + typeOfSchema(schema), processed.has(schema) ? '(FROM CACHE)' : '')
-
+  let scopedMakeName = () => { return makeName(options, schema, usedNames, keyName, keyNameFromDefinition) }
```
bcherny: Can you avoid the refactor here, and just add your new logic to the existing `standaloneName` function (modifying `standaloneName` to also take `options`)?

src/utils.ts
```diff
@@ -88,6 +89,83 @@ export function generateName(from: string, usedNames: Set<string>) {
...
+  keyName?: string
+  /**
+   * this is the key of this schema object under definitions if applicable, this may be the same as `keyName`
+   */
+  definitionKeyName?: string
+  /**
+   * this is a set of all names that have already been used
+   */
+  usedNames: Set<string>
+}
```
bcherny: No need for named arguments. Just make `customName` take 4 arguments. Let's also make sure that `customName` returns a `string` (rather than `string | undefined`).

bcherny:

PR #219, opened 2019-02-18
Escape comments

SimonSiefke:
Escape comments like /* */ in description properties


src/normalizer.ts
```diff
@@ -43,6 +43,30 @@ rules.set('Default top level `id`', (schema, rootSchema, fileName) => {
...
+  replacer = '*\\/'
+) {
+  if (code === null || typeof code !== 'object') {
+    return
+  }
+  for (const key of Object.keys(code)) {
+    if (key === 'description' && typeof code[key] === 'string') {
+      code[key] = code[key].replace(/\*\//g, replacer)
+    } else {
+      escape(code[key])
```
bcherny: I don't think you need to manually recurse. IIRC `mapDeep` handles this for you.

src/normalizer.ts
```diff
@@ -43,6 +43,30 @@ rules.set('Default top level `id`', (schema, rootSchema, fileName) => {
   return schema
 })
 
+function escape(
+  code: string | { [key: string]: any },
```
bcherny: Call this `schema`, to align with the other rules. Also, its type should just be `JSONSchema`.

src/normalizer.ts
```diff
@@ -43,6 +43,30 @@ rules.set('Default top level `id`', (schema, rootSchema, fileName) => {
   return schema
 })
 
+function escape(
```
bcherny: Call this `escapeBlockComment`, and move it to *utils.ts*.

SimonSiefke:
Can you take another look? :)

bcherny:
Published `6.1.2`

PR #195, opened 2018-10-17
Add /* tslint:disable */ to the default banner

deadNightTiger:
Fixes #192

PR #221, opened 2019-03-26
Add /* prettier-ignore */ to the default banner

kalitine:
By analogy with #195 Pull Request


krokofant:
Looks good to me! Since most Node.js files are written in JavaScript I would (and have locally) add a eslint disable flag as well. 

```javascript
/* eslint-disable */
```

krokofant:
@bcherny Yes, but as I see it there are three options:
1. Use the formatting rules from the consumers prettier config (where they can also specify ignores)
2. Ignore all formatting all together
3. Apply the banners _after the formatting is done_

I like the third option so this tool can have decent defaults but exclude it from formatting and style checks on the consumer end.

bcherny:

PR #226, opened 2019-04-04
Moved unnecessary @types to devDependencies

devoto13:
`cli-color`, `lodash`, `minimist` and `mz` are not exported by the library, so consumers don't need to install their types. 

Also changed `@types/node` to `*` to ensure that it is de-duped by the package manage if consumer already has this dependency (in particular older version).

These changes result in a more lightweight installation (`@types` folder used to be 2.2MB, now 594KB).

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,12 +45,8 @@
   },
   "homepage": "https://github.com/bcherny/json-schema-to-typescript#readme",
   "dependencies": {
-    "@types/cli-color": "^0.3.29",
     "@types/json-schema": "^7.0.3",
-    "@types/lodash": "^4.14.121",
-    "@types/minimist": "^1.2.0",
-    "@types/mz": "0.0.32",
-    "@types/node": "^11.10.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -45,12 +45,8 @@
   },
   "homepage": "https://github.com/bcherny/json-schema-to-typescript#readme",
   "dependencies": {
-    "@types/cli-color": "^0.3.29",
     "@types/json-schema": "^7.0.3",
-    "@types/lodash": "^4.14.121",
-    "@types/minimist": "^1.2.0",
-    "@types/mz": "0.0.32",
-    "@types/node": "^11.10.4",
+    "@types/node": "*",
```
bcherny: Could we add a lower bound here, eg. `>=6.0.0` or something reasonable?

devoto13: Set to `>= 4.5.0`, same as `engines` property.

HoldYourWaffle:
Is there a reason why this hasn't been merged yet?

Issue #131, opened 2017-11-14
Add option for generate interface with readonly properties


ikokostya:
**The problem**

I want to generate interface for object which must not be changed. I can manually wrap result using `DeepReadonly` type:

```ts
type DeepReadonly = {
   readonly [P in keyof T]: DeepReadonly<T[P]>;
};
```
but it doesn't work with arrays:
* https://github.com/Microsoft/TypeScript/issues/10725
* https://github.com/Microsoft/TypeScript/issues/13923

**Suggestion**

Add some option to `json-schema-to-typescript` for generating readonly properties. For example from `README.md`:

```ts
export interface ExampleSchema {
  readonly firstName: string;
  readonly lastName: string;
  readonly age?: number;
  readonly hairColor?: "black" | "brown" | "blue";
}
```

For arrays `ReadonlyArray` type can be used:

```ts
export interface ExampleSchema {
    readonly arr: ReadonlyArray<string>
}
```

**Questions**

Unfortunately there is no way to create read only tuple (as I known) if `json-schema-to-typescript` produces it.

bcherny:
That's a really great idea @ikokostya! We might get requests to split it into two settings at some point (readOnlyProperties via readonly keyword, readOnlyValues via Readonly<> or ReadonlyArray<> types) - what do you think of splitting it that way?

Anyway, PRs are welcome!

PR #230, opened 2019-05-02
Add readonly support

mattroberts297:
Because I wanted to enforce some immutability

_PS Thanks for the awesome library / tool_


jameswilddev:
This looks like it would resolve #131, excepting the use of ReadOnlyArray.  Do you know if it would also turn `{ [k: string]: T }` into `{ readonly [k: string]: T}`?

bcherny:
Looks good, but I'd love if you could iterate on this a bit more before we merge it (also, sorry for the delay in getting to this!).

1. Please rename the option to `--readonly`.
2. Could you make sure it also makes index signatures (as @jameswilddev noted), array types, tuple types, and referenced types readonly?
3. Please also update README.md to document this.

jonaskello:
I'm also looking for this function, is someone still working on it?

mattroberts297:
I stopped working on this PR because I moved away from the language (as opposed to the library) for a time. Thanks for the feedback and apologies for not addressing it.

codingismy11to7:
I'm guessing this PR will probably get closed, but just wanted to chime in that I'd additionally expect
```typescript
export type X = Base & {
  field: type
};
```

to output as 
```typescript
export type X = Base & Readonly<{
  field: type
}>;
```

which this code doesn't seem to address

bcherny:
https://github.com/bcherny/json-schema-to-typescript/pull/412

PR #232, opened 2019-05-24
Move @types packages to devDependencies

ikokostya:
Moved all `@types` packages to `devDependencies`.  

Refs #226

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,13 +45,6 @@
   },
   "homepage": "https://github.com/bcherny/json-schema-to-typescript#readme",
   "dependencies": {
-    "@types/cli-color": "^0.3.29",
-    "@types/json-schema": "^7.0.3",
-    "@types/lodash": "^4.14.121",
-    "@types/minimist": "^1.2.0",
-    "@types/mz": "0.0.32",
-    "@types/node": "^11.10.4",
-    "@types/prettier": "^1.16.1",
     "cli-color": "^1.4.0",
     "json-schema-ref-parser": "^6.1.0",
     "json-stringify-safe": "^5.0.1",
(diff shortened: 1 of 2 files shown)


devoto13:
Moving all packages is wrong and will break consumers of the json-schema-to-typescript as it exports types defined in some of these libraries. And then it is essentially a dupe of https://github.com/bcherny/json-schema-to-typescript/pull/226.

ikokostya:
> Moving all packages is wrong and will break consumers of the json-schema-to-typescript as it exports types defined in some of these libraries. 

Why it is wrong? It is common practice to not include types of external packages to dependencies. 

If it is breaking change, major version should be bumped.

devoto13:
@ikokostya It is a common practice to exclude *unnecessary* types. Otherwise you won't be able to use the package, because it fails type checking.

I would suggest to do something like below to see what the problem is:

```
$ mkdir fff
$ cd fff/
$ npm init -f
$ npm install --save typescript json-schema-to-typescript
$ touch index.ts
import { compile } from 'json-schema-to-typescript'
compile(null, 'MyScheam')
$ ./node_modules/.bin/tsc index.ts # compiles successfully
$ rm -rf node_modules/@types/
$ ./node_modules/.bin/tsc index.ts # here are the errors consumers will get with the changes from this PR
```

qzb:
@ikokostya, [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies) explains pretty clearly why (and when) types have to be added to dependencies:

> (...) Our package exposes declarations from each of those, so any user of our `browserify-typescript-extension` package needs to have these dependencies as well. For that reason, we used `"dependencies"` and not `"devDependencies"`, because otherwise our consumers would have needed to manually install those packages. If we had just written a command line application and not expected our package to be used as a library, we might have used `devDependencies`.

bcherny:
Dup of #226

PR #238, opened 2019-06-18
add cli directory support (using globs)

Ethan-Arrowood:
I'm working on adding cli support for glob patterns and directories of JSON files. This probably isn't complete just yet but I thought I'd share my work so I could get some initial feedback.

Ref https://github.com/bcherny/json-schema-to-typescript/issues/16

I need to familiarize myself with the library more, but it seems like there is need to add this to the programmatic side too (https://github.com/eweap/json-schema-to-typescript/blob/97f91e758317ef3ea5be3b9a1dfdd162051e3a48/src/index.ts#L68-L92). I'll try and use that as a reference


Ethan-Arrowood:
FWIW this may also be better implemented in another library. Something that wraps this one. I'll let the maintainers decide what they'd like best

src/cli.ts
```diff
@@ -23,18 +29,41 @@ async function main(argv: minimist.ParsedArgs) {
...
+      await writeOutput(ts, argOut)
+    }
   } catch (e) {
     console.error(whiteBright.bgRedBright('error'), e)
     process.exit(1)
   }
+}
 
+function processFiles(files: string[], outDir: string, argv: Partial<Options>) {
+  return new Promise(async (res, rej) => {
```
bcherny: Could you use `Promise.all` here for better concurrency?

src/cli.ts
```diff
@@ -3,16 +3,22 @@
...
+const glob = promisify(_glob)
+
 main(minimist(process.argv.slice(2), {
   alias: {
     help: ['h'],
     input: ['i'],
-    output: ['o']
+    output: ['o'],
+    inDir: ['x'],
+    outDir: ['y']
```
bcherny: Would it work if we reuse the existing `i` and `o` args, and avoid introducing `x` and `y`?

src/cli.ts
Ethan-Arrowood: I'll try this! Thank you for the recommendation

src/cli.ts
Ethan-Arrowood: Yes probably a good idea!

Ethan-Arrowood:
Okay made changes; working on testing now

Ethan-Arrowood:
Haha testing was a bit of a challenge but I got something working! Let me know what else you'd like for this new feature :)

.vscode/settings.json
```diff
@@ -1,5 +1,19 @@
 // Place your settings in this file to overwrite default and user settings.
 {
     "typescript.tsdk": "node_modules/typescript/lib",
-    "editor.tabSize": 2
+    "editor.tabSize": 2,
+    "workbench.colorCustomizations": {
```
bcherny: Please remove the changes to this file from this PR.

src/cli.ts
```diff
@@ -27,14 +32,37 @@ async function main(argv: minimist.ParsedArgs) {
   const argOut: string = argv._[1] || argv.output
 
   try {
-    const schema: JSONSchema4 = JSON.parse(await readInput(argIn))
-    const ts = await compile(schema, argIn, argv as Partial<Options>)
-    await writeOutput(ts, argOut)
+    if (isGlob(argIn)) {
+      if (!existsSync(argOut)) {
+        await mkdir(argOut)
```
bcherny: What if the dir is nested (`/foo/bar/baz`) -- should we use `mkdirp` here?

src/cli.ts
```diff
@@ -27,14 +32,37 @@ async function main(argv: minimist.ParsedArgs) {
...
+      }
+
+      const files = await glob(join(process.cwd(), argIn))
+      const processFiles = files.map(file => processFile(file, { dir: argOut }, argv as Partial<Options>))
```
bcherny: Could we DRY this up a bit? One option is something like:

```ts
try {
  await Promise.all(getFilesToProcess(argIn));
} ...

function getFilesToProcess(argIn: string): Promise<void>[] {
  if (isGlob(....)) {
    ...
    return files.map(...)
  }
  return [processFile(...)]
}
```

test/testCLI.ts
```diff
@@ -82,4 +82,15 @@ export function run() {
     unlinkSync('./ReferencedType.d.ts')
   })
 
+  test('files in (-i), files out (-o)', t => {
```
bcherny: This looks great! Would you mind also adding test cases for:

- Input directory specified, output directory not specified
- Input directory specified, output directory is specified and nested, and does not exist (/foo/bar/baz)
- Input directory is specified,

src/cli.ts
```diff
@@ -27,14 +32,37 @@ async function main(argv: minimist.ParsedArgs) {
...
   try {
-    const schema: JSONSchema4 = JSON.parse(await readInput(argIn))
-    const ts = await compile(schema, argIn, argv as Partial<Options>)
-    await writeOutput(ts, argOut)
+    if (isGlob(argIn)) {
+      if (!existsSync(argOut)) {
+        await mkdir(argOut)
+      }
+
+      const files = await glob(join(process.cwd(), argIn))
```
bcherny: Could we move this up (above creating the output directory), and short-circuit early with a nice error message when `files.length === 0` (something like "Couldn't find any files using the glob [glob].")?

.vscode/settings.json
Ethan-Arrowood: My apologies! I didn't realize my editor modified this; will revert.

Ethan-Arrowood:
Thank you for the review; will work on these recommendations soon and push changes!

Ethan-Arrowood:
sorry it has been a little while, got swept up with final projects (yay for still being in college).

Ethan-Arrowood:
The test suite is passing but I'm getting unhandelled promise rejection errors in the console. Trying to track it down now.

Ethan-Arrowood:
@bcherny hey! I'd like to try and complete this PR. I tried rebasing on master and I seem to have messed up the snapshots. Can you advise how I can fix this?

Ethan-Arrowood:
For those looking for directory access I've now published another npm module that implements such https://www.npmjs.com/package/compile-schemas-to-typescript (only api no cli right now)

bcherny:
@Ethan-Arrowood Hey! Missing your comment.

You want to re-generate the snapshots:

```sh
git rebase origin/master
yarn test -u
```

Ethan-Arrowood:
Okay rebased and updated but there is some dep issue? Not sure whats wrong there

src/cli.ts
```diff
@@ -28,15 +43,53 @@ async function main(argv: minimist.ParsedArgs) {
...
+
+        if (files.length === 0) {
+          rej('No files match glob pattern')
+        }
+
+        if (argOut && !existsSync(argOut)) {
+          await mkdirp(argOut)
+        }
+
+        res(files.map(file => processFile(file, {dir: argOut}, argv)))
```
bcherny: Could we make this a little more explicit, and kill some of the boilerplate?

1. Rename `getFilesToProcess` to `processFiles`, update its return type to `Promise<void[]>`, and mark it `async`
2. Remove the `Promise.all` on 47
3. Remove `processFiles`'s explicit `new Promise...` wrapper (lines 55, 56, 73-77)
4. Update L68 to:

```ts
  return Promise.all(
    files.map(file => processFile(file, {dir: argOut}, argv))
  )
```

Also:

1. Can we also remove the duplication between line 68 and 71, by setting `files` to `[argIn]` early when the input is not a directory or a glob?
2. What will happen when the input glob resolves to a file tree with multiple sub-folders that need to be `mkdirp`'d? Can we make sure this works, and add a test case for it? eg.

```
/input
  /a
    /x.json
  /b
    /y.json
/output
```

Where `/output` does not have any sub-folders, we should create both `/output/a` and `/output/b`.

test/testCLI.ts
```diff
@@ -78,4 +78,34 @@ export function run() {
     t.snapshot(readFileSync('./ReferencedType.d.ts', 'utf-8'))
     unlinkSync('./ReferencedType.d.ts')
   })
+
+  test('files in (-i), files out (-o)', t => {
+    execSync(
+      'node dist/src/cli.js -i ./test/resources/MultiSchema/**/*.json -o ./test/resources/MultiSchema/out'
+    ).toString()
```
bcherny: Looks like you're mimicking the old test cases here, but I think you can safely remove the `.toString()` here and elsewhere right?

test/testCLI.ts
```diff
@@ -78,4 +78,34 @@ export function run() {
     t.snapshot(readFileSync('./ReferencedType.d.ts', 'utf-8'))
     unlinkSync('./ReferencedType.d.ts')
   })
+
+  test('files in (-i), files out (-o)', t => {
```
bcherny: Should we add a test case that we concat all generated TS into one file when the input is a directory, while the output is a filename? eg. `-i a/**/*.json -o b.json` should concatenate all generated TS into `b.json`.

Ethan-Arrowood: I considered this feature case out of scope for this PR. If you think otherwise I can try to implement it

src/cli.ts
Ethan-Arrowood: I like this idea. I will try and add this next

Ethan-Arrowood:
Working on adding support for nested directory output now

test/testCLI.ts
bcherny: It seems fair to me to leave it out of scope. If you choose to do that, would you mind at least detecting that case and throwing a `ReferenceError` when we hit it?

src/cli.ts
```diff
@@ -25,26 +40,50 @@ async function main(argv: minimist.ParsedArgs) {
...
+
+  if (argOut && extname(argOut) === '' && !existsSync(argOut)) {
+    await mkdirp(argOut)
+  }
+
+  return Promise.all(files.map(file => processFile(file, argOut, argv)))
+}
+
+async function processFile(file: string, out: string | undefined, argv: Partial<Options>): Promise<void> {
+  out = out ? join(process.cwd(), out) : ''
```
bcherny: What does it mean for `out` to be `''` here? What will the behavior be (and could we make that behavior more explicit?

Also:

- It might be worth killing `writeOutput`, and splitting it into `writeOutputToStdOut` and a regular `writeFile` call
- It would be nice to avoid querying the filesystem for whether `out` is a directory or not. Instead, we should always pass a file path to `processFile`.

test/testCLI.ts
Ethan-Arrowood: I'll also open a new issue to track that feature request so others could contribute it 😄

src/cli.ts
```diff
@@ -1,13 +1,27 @@
...
+import * as _mkdirp from 'mkdirp'
+import * as _glob from 'glob'
+import isGlob = require('is-glob')
+import {promisify} from 'util'
+import {join, resolve, dirname, basename} from 'path'
 import stdin = require('stdin')
 import {compile, Options} from './index'
 
+// Promisify mkdirp & glob
+const mkdirp = (path: string) =>
```
edzis: I would consider using `fsPromises.mkdir(path, {recursive: true})` instead - see https://nodejs.org/api/fs.html#fs_fspromises_mkdir_path_options

Ethan-Arrowood: if I remember correctly that option only exists on node version 10+ but this lib needs to support node 8 so I wasn't able to use it

Ethan-Arrowood:
Fixing tests and adding the single output error catch now 😄

Ethan-Arrowood:
So I added the error check, but struggled to write a test for it. execSync isn't processing the error correctly for ava .throws() to catch

Ethan-Arrowood:
@bcherny can you confirm what version of node this library is supposed to support? The package.json says 4.5 but I'm having issues with the @types/node version being so out of date. My latest commit bumps it to latest let me know if I should revert that and use old-style node code instead

test/testCLI.ts
MaikuMori: Simple concatenation doesn't work. For example if you have 3 schemas.

> A, B, C
> A refers definition in C
> B refers definition in C

In the output you'll have 3 files defining the same definition from C. If you just concatenate files, you'll get duplicate declaration errors.

Many to one file support would be very handy.

bcherny: That’s a really good point — though TS will merge interfaces for us, it will error when multiple files emit type aliases with the same name.

To fix that, we’ll need a way to de-dupe across files, ideally.

Ethan-Arrowood:
Bumping - what are my next steps to get this completed?

package.json
```diff
@@ -45,20 +45,26 @@
   },
   "homepage": "https://github.com/bcherny/json-schema-to-typescript#readme",
   "dependencies": {
+    "@types/is-glob": "^4.0.1",
     "@types/json-schema": "^7.0.3",
-    "@types/node": ">=4.5.0",
+    "@types/mkdirp": "^0.5.2",
+    "@types/node": "^13.1.2",
```
bcherny: Any chance we could make this a bit more permissive, so it works with peoples' existing `@types/node`? eg. `>=10` or `>=11` would be nice, since 12 is LTS.

src/cli.ts
bcherny: Any chance you could add an explicit return type here (`Promise<???>`)? TypeScript isn't smart enough to infer it correctly.

src/cli.ts
```diff
@@ -25,35 +39,120 @@ async function main(argv: minimist.ParsedArgs) {
...
   const argIn: string = argv._[0] || argv.input
-  const argOut: string = argv._[1] || argv.output
+  const argOut: string | undefined = argv._[1] || argv.output // the output can be omitted so this can be undefined
+
+  const ISGLOB = isGlob(argIn)
+  const ISDIR = isDir(argIn)
+
+  if ((ISGLOB || ISDIR) && argOut && argOut.includes('.d.ts')) {
+    throw new ReferenceError(
+      `You have specified a single file ${argOut} output for a multi file input ${argIn}. This feature is not yet supported, refer to issue #272 (https://github.com/bcherny/json-schema-to-typescript/issues/272)`
```
bcherny: 👍

src/cli.ts
```diff
@@ -25,35 +39,120 @@ async function main(argv: minimist.ParsedArgs) {
...
-  return readFile(resolve(process.cwd(), argIn), 'utf-8')
+  // create output directory if it does not exist
+  if (argOut && !existsSync(argOut)) {
+    await mkdirp(argOut)
+  }
+
+  return Promise.all(
+    files.map(file => {
+      const outPath = argOut ? `${argOut}/${basename(file, '.json')}.d.ts` : argOut
+      processFile(file, outPath, argv)
```
bcherny: Do you need to `return` this?

src/cli.ts
```diff
@@ -25,35 +39,120 @@ async function main(argv: minimist.ParsedArgs) {
...
   }
-  return readFile(resolve(process.cwd(), argIn), 'utf-8')
+  // create output directory if it does not exist
+  if (argOut && !existsSync(argOut)) {
+    await mkdirp(argOut)
+  }
+
+  return Promise.all(
+    files.map(file => {
+      const outPath = argOut ? `${argOut}/${basename(file, '.json')}.d.ts` : argOut
```
bcherny: Nit: To make this a little more understandable, could you replace `: argOut` with `: undefined`?

src/cli.ts
```diff
@@ -25,35 +39,120 @@ async function main(argv: minimist.ParsedArgs) {
...
+
   if (!argOut) {
-    try {
-      process.stdout.write(ts)
-      return Promise.resolve()
-    } catch (err) {
-      return Promise.reject(err)
-    }
+    process.stdout.write(ts)
+    return
```
bcherny: This `return` seems redundant

src/cli.ts
```diff
@@ -25,35 +39,120 @@ async function main(argv: minimist.ParsedArgs) {
...
+      if (!argOut) {
+        processFile(file, argOut, argv)
+      } else {
+        /*
+        the following logic determines the out path by comparing the in path to the users specified out path.
+        For example, if input directory MultiSchema looks like:
+        MultiSchema/foo/a.json
+        MultiSchema/bar/fuzz/c.json
+        MultiSchema/bar/d.json
+        And the user wants the outputs to be in MultiSchema/Out, then this code will be able to map the inner directories foo, bar, and fuzz into the intended Out directory.
```
bcherny: This comment is helpful! Mind extending this comment to also include what we'll map these to?

ie.

```
MultiSchema/Out/foo/a.json
MultiSchema/Out/bar/fuzz/c.json
MultiSchema/Out/bar/d.json
```

src/cli.ts
```diff
@@ -25,35 +39,120 @@ async function main(argv: minimist.ParsedArgs) {
...
+        */
+        const outPathList = argOut.split('/')
+        const inPathList = file.split('/')
+
+        const intersection = outPathList.filter(x => inPathList.includes(x))
+        const difference = outPathList
+          .filter(x => !inPathList.includes(x))
+          .concat(inPathList.filter(x => !outPathList.includes(x)))
+
+        let outPath = join(...intersection, ...difference)
```
bcherny: Would you mind pulling lines 110-118 out into its own function, and writing some unit tests for it? I'm having trouble understanding if this will work right in all cases.

Ethan-Arrowood:
Sorry for being MIA - will get to this asap (work has picked up significantly)

src/cli.ts
Ethan-Arrowood: Maybe, I'll see what happens without it

src/cli.ts
Ethan-Arrowood: I believe I can replace this with `argOut || ''` instead

src/cli.ts
Ethan-Arrowood: It uses set theory. Not sure if unit testing would be best here, but happy to provide some if that is what you think is best.

Ethan-Arrowood: The _proof_ for this is:

The first two lines split the file path into lists separated by the `/`.
```
const outPathList = o.split('/')
const inPathList = i.split('/')
```

The next line creates an intersection set of the two lists. This will contain only the parts of the file path that are shared. 
```
const intersection = outPathList.filter(x => inPathList.includes(x))
```

Next, we create a symmetric difference of the out path. This is essentially the list of paths that are not shared between the two paths.
```
  const difference = outPathList
    .filter(x => !inPathList.includes(x))
    .concat(inPathList.filter(x => !outPathList.includes(x)))
```

Then we join these two lists together (with the intersection list first) to produce the transformed output.
```
return join(...intersection, ...difference)
```
---
An example, consider the following input directory `MultiSchema`
```
MultiSchema/foo/a.json
MultiSchema/bar/fuzz/c.json
MultiSchema/bar/d.json
```
And the desired output directory is `MultiSchema/Out`
Then the CLI will perform the following function calls:
```
pathTransform('MultiSchema/Out', 'MultiSchema/foo/a.json')
pathTransform('MultiSchema/Out', 'MultiSchema/bar/fuzz/c.json')
pathTransform('MultiSchema/Out', 'MultiSchema/bar/d.json')
```
And will result in the intended output:
```
'MultiSchema/Out/foo/a.json'
'MultiSchema/Out/bar/d.json'
'MultiSchema/Out/bar/fuzz/c.json'
```

bcherny: A proof is great, tests are better, if that's not too much to ask.

Even Haskellers use QuickCheck :)

src/cli.ts
```diff
@@ -25,35 +39,135 @@ async function main(argv: minimist.ParsedArgs) {
...
+        processFile(file, argOut, argv)
+      } else {
+        let outPath = pathTransform(argOut, file)
+        // const outPathList = argOut.split('/')
+        // const inPathList = file.split('/')
+
+        // const intersection = outPathList.filter(x => inPathList.includes(x))
+        // const difference = outPathList
+        //   .filter(x => !inPathList.includes(x))
+        //   .concat(inPathList.filter(x => !outPathList.includes(x)))
```
bcherny: Are these intentionally commented out?

Ethan-Arrowood: Ahh oops meant to delete after i put it into its own function. Fixing now (and adding tests 😉)

Ethan-Arrowood:
Okay so I added a test, but ava is failing with a nondescript error:

```
> ava --serial --verbose


error [Error: ENOENT: no such file or directory, open '/Users/ethanarrowood/Documents/Programming/json-schema-to-typescript/undefined'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/ethanarrowood/Documents/Programming/json-schema-to-typescript/undefined'
}
  ✖ dist/test/test.js exited with a non-zero exit code: 1
```
Do you have any ideas what I'm doing wrong in my code? I'm sorry I'm not familiar with Ava at all 😕

bcherny:
@Ethan-Arrowood Which directory are you running that from? Try importing your test file from https://github.com/bcherny/json-schema-to-typescript/blob/master/test/test.ts, then running `yarn test`

Ethan-Arrowood:
Project root, using `npm run test` and the test file is already included (I added the test to the end of the cli tests considering the function is related to the cli code). Do I need to extrapolate the function into its own file? And similarly the test code into its own file too?

bcherny:
Want to push the commit? Happy to pull it to help debug.

Ethan-Arrowood:
Yep! Here it is: https://github.com/bcherny/json-schema-to-typescript/pull/238/commits/5c1e875cfa889c5f971b8c675ce8ff74772087a6

bcherny:
@Ethan-Arrowood So you want to pull your `pathTransform` function into utils.ts, and make a new test file for it -- testUtils.ts. Otherwise, when you import `pathTransform` from cli.ts, you're accidentally triggering `main()` in cli.ts, which is throwing your error.

Ethan-Arrowood:
I think its ready to go!

bcherny:
@Ethan-Arrowood Mind rebasing?

Ethan-Arrowood:
rebased :D

src/cli.ts
bcherny: Did we address this? ^

src/cli.ts
```diff
@@ -25,35 +40,103 @@ async function main(argv: minimist.ParsedArgs) {
...
 }
 
-function writeOutput(ts: string, argOut: string): Promise<void> {
+async function processDir(argIn: string, argOut: string | undefined, argv: Partial<Options>): Promise<void[]> {
+  const files = getPaths(argIn)
+
+  return Promise.all(
+    files.map(file => {
+      if (!argOut) {
+        processFile(file, argOut, argv)
```
bcherny: And here

src/cli.ts
```diff
@@ -25,35 +40,103 @@ async function main(argv: minimist.ParsedArgs) {
...
+      } else {
+        let outPath = pathTransform(argOut, file)
+        if (!isDir(dirname(outPath))) {
+          _mkdirp.sync(dirname(outPath))
+        }
+        outPath = outPath.replace('.json', '.d.ts')
+        processFile(file, outPath, argv)
```
bcherny: And here

src/cli.ts
Ethan-Arrowood: Strange, yes and the change is present when I jump over to file view.

src/cli.ts
Ethan-Arrowood: Are you referencing the `return Promise.all`? I will remove this and make sure it still works

Ethan-Arrowood: What change are you expecting here?

src/cli.ts
Ethan-Arrowood: What change are you expecting here?

Ethan-Arrowood:
And yes I can update the README!

src/cli.ts
bcherny: It’s been a while since I’ve touched JS concurrency, but are promises actually executed concurrently, with the Promise.all revolving when the longest promise resolves, if you don’t return them in all these places?

Ethan-Arrowood: I sadly do not know. Async/await behavior and promises are confusing for me so I don't know exactly what would happen. Assuming the mdn docs are true [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) does resolve when _all_ promises resolve.

bcherny:
Published 8.2.0.

Ethan-Arrowood:
Wooo so happy this got merged! I'll be sure to contribute more in the future 😄

Issue #218, opened 2019-02-01
Tuple items with `minItems` and `additionalItems`


epferrari:
We have a use case I don't think is currently covered. We have an array where 2 items must be defined, a max of 3, and the first one has an additional constraint. JsonSchema allows for this, but the generated type isn't what we'd expect.

schema:
```
{
  "title": "Foo",
  "type": "array",
  "minItems": 2,
  "maxItems": 3,
  "definitions": {
    "foo_0": {
      "title": "Foo Item",
      "type": "integer",
      "exclusiveMinimum": 0
    },
    "foo_n": {
      "title": "Foo Item",
      "type": "integer"
    }
  },
  "items": [{
    "$ref": "#/definitions/foo_0"
  }, {
    "$ref": "#/definitions/foo_n"
  }],
  "additionalItems": {
    "$ref": "#/definitions/foo_n"
  }
}
```

expected generated ts:
```
export type Foo = [FooItem, FooItem1, FooItem1?];
export type FooItem = number;
export type FooItem1 = number;
```

actual generated ts:
```
export type Foo = [FooItem, FooItem1];
export type FooItem = number;
export type FooItem1 = number;
```

using version 6.1.1

bcherny:
This is a fun one :)

- Bug #1 is when we infer a tuple type, we should look to `additionalItems` as the spread property. In your example, the correct type to generate is `[FooItem, FooItem1, ...FooItem1[]]`.
- Enhancement #2 is we should take `minLength` and `maxLength` into account when specifying tuple types. As a first pass:
  1. When minLength !== items.length, we should throw a validation error
  2. When maxLength is defined, we should generate optional tuple properties (eg. `FooItem1?`) instead of a rest property (eg. `...FooItem1`).

Want to take a shot at Bug 1?

PR #241, opened 2019-06-29
 feat: support `additionalItems` on tuple types

bradzacher:
Fixes half of #218 

There's a few unrelated lint fixes that I noticed were failing (separated in b8bf6ed)


src/generator.ts
```diff
@@ -158,9 +166,21 @@ function generateType(ast: AST, options: Options): string {
...
     case 'REFERENCE': return ast.params
     case 'STRING': return 'string'
-    case 'TUPLE': return '['
-      + ast.params.map(_ => generateType(_, options)).join(', ')
+    case 'TUPLE': return (() => {
+      const params = ast.params.map(_ => generateType(_, options))
+      if (ast.spreadParam) {
+        let spread = generateType(ast.spreadParam, options)
+        if (spread.endsWith('"')) {
+          spread = '(' + spread + ')'
```
bcherny: Nit: Consider always adding the `()` for simplicity, and letting the prettifier strip them out for you if they're not necessary.

src/generator.ts
```diff
@@ -82,6 +86,9 @@ function declareNamedInterfaces(
     case 'TUPLE':
     case 'UNION':
       type = ast.params.map(_ => declareNamedInterfaces(_, options, rootASTName, processed)).filter(Boolean).join('\n')
+      if ('spreadParam' in ast && ast.spreadParam) {
```
bcherny: Nit: The second check seems to be redundant, since you defined the member type to be non-nullable.

```diff
-if ('spreadParam' in ast && ast.spreadParam) {
+if ('spreadParam' in ast) {
```

test/__snapshots__/test/test.ts.md
```diff
@@ -279,6 +279,153 @@ Generated by [AVA](https://ava.li).
...
+    `/* tslint:disable */␊
+    /**␊
+     * This file was automatically generated by json-schema-to-typescript.␊
+     * DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,␊
+     * and run json-schema-to-typescript to regenerate this file.␊
+     */␊
+    ␊
+    export interface ArrayAdditionalItems {␊
+      namedEums?: {␊
+        additionalItemsAny?: [NamedEnum2, ...any[]];␊
```
bcherny: I wonder if this should be `...unknown[]` instead. Or maybe we should add a `--strictNoAny` flag in the future, so that all generated `any`s (not just this one) are `unknown` instead.

src/generator.ts
bradzacher: Don't confuse TS with flow!

in TS w/ strict mode
`spreadParam?: AST` is the same as
`spreadParam?: AST | undefined`

test/__snapshots__/test/test.ts.md
bradzacher: I'd probably lean toward the later solution I think, just so that the output is consistent across the package (it seems weird to introduce a single `unknown` generation point).

src/generator.ts
bradzacher: Interestingly, prettier doesn't actually remove them if they're completely unnecessary.

[prettier repl](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACAS1jmwDMBDMOPAQW23IE8aATFgmA6cgGwEl4AWwDOeYAB0oeaTKjlBcFgFEAriID8SMZJm685Nhy5z+Q4TSiNNeANoA5eYqVQ1AJgA0eAHQ+AFOUsAShsAXRCAbh09aQN2Tm5TOA0te0dlF0FXAEZPHy9fBwV0t1dgsMipPQBfCuiVKDkipUE1YWsJSujYowSBJPNLaxtfLLwAHzxXcbwAZkDcvwDGMoiorsN4kz7k219xEAB5KDh96f2AFQB3CFOJi4ALbDgTkHnvPwAWaYBWaYA2Fa1apA3SNRQAZTA9yS5Haaz03U2vG2AysKXBMGwRAA5pDoYJyDk8HY1AAjEh4mFEvL+IKhVadBEbYzIsxDDFYqC4qEwjzEskUnkEvk0gBCEAgPDgAUpBKygPh0hqirwwiFsK0HWiMWZvTMFjRthVui12t0pAl1mEmJxILNtgA1lprZzsSEtEs7dEqu5jTJTfa8KTyNhrBlydgvdqbE7VTaue79JYo7ofX7pDSlmU-QyzYiWYkdjZ09pGWaLRArfHsSm9DHndXE56S2my3oA-bg6GtOGSLXdPW466m8mW7627oaR2zcGAF7WUkSqUBfsyQcunEjxirqrZid4XPAlUQUjnRgYOBw-eIvX9A1DXvYTwbrkLfJZ+m1m9bNkpR-P6s318RdJWlKAFTbZU2yeUg2k1P181vER7xSJQ0HkDApVlcggI-coEN1H9+iGNCMKw9VqT8UjBEwuBsNKT8VSg6IWDgWCrzzQjWTvQYUgAMQIbBrQAETYogeigXC6Xw68uMLODbAAVSgJ4eHIeAWHODgpSA857kElhRNIcTNgg707SqeF0CwXBCGIMhKGoDkcWw0ZpzwCsq1dKN10bD1R0ZSzGWsnB8CIeAHKoflBAjVzS2iLswwFSMVV84d-O3eEgt0ELbPCkgKCi5yuXo+K9E8htvNS2MXzdDKLKszBQrsiLCuoEkYsFfFyCmdzEp7ZKfJqvyk0ywLGpssL7LavBxVAmUKLK3Q5wXJcwKGyrN3qrKJua-LIuoajaOwpaZGMoSYEKOBNq5KM1OtK6bprFUAHoACo3uNN7aGxagiDwRhpSEr6XpVchfqSzqUrbaECGxe4YEhiMowoAA3HAODgPiJRYBSllCO6CAdOBhGEiBsQUkDlygDahy20aGuCpq8umxy8CO8jurclVzoexwnru8g+YUAXXo+r6fr+qRAZDYQQbBiGBqhqNYfhxGleRnnyHRrF4GxiBcesfHD10HgiZJsmKdW+aaeqp6t0ZnLmam1q2Y5uj1V6lUCGEABhaBSDNsAYC0Kn1p2xl3s+stvq0mApTwAkYB5URWOMqAJLwYnGA8nAs4IMAHTlmPQaZyaWoKtmBIuoyTOMU7pDBLzbQj3Qo8VWPtL+0R6lU9TFDwGAIDwNO6+gLO4BzgIWCDOB7m1rgVGwQhRBgaFVUcDvS6d8v9pm5S+40uOE-cpvRfGyPxZjvBa4zzY8Hn0QoGHzh4+oae8AwbADZUKgn+HlgKhMIF37ngPgrMqBb12izV2UU9IGVvpnU+jhm63VbjIaB+BIBQGtHgBAahiRpGcG4BueAjjUAALx4ByCqK4w8qEeFoY8Z4eAqEzHQdIXKWDoC4PwYIQhTQMjZFIeQ1h1DxzRDoWIxhbY9JPEobMDheDnZ4GwbwjIAinBCK9m2bGS8xEfAkXoASqMFHfCMbocEBA0BiL+FlEA7gQAQAwJsYQyBQAhm-pcAACiGBAbiUDawgAQFgDiQCkgYIXOAMBwQYEoDiZAmIVBwEcQjQQPAADq+l4DCDiVQcE-ijCow4IwZA4BhBuMcUQYQJAYDeIYNiAkyAKA8BqY4gAVsINAopInExiY4AAMkQOAzTeBtJAJ0tAViuRSgAIoqAgPAUZrSUkgDiUJEgZSYDnhJmALELiwlf3ChkkJa9kAAA4AAMjiv4QBqRkhgGAylfxJiQUxYSACOCz4D1OcQEkAQsAC0xxFCKDCU8L5gk4D1PBk0pALTxk1MEAQRJ2BkmOOEDiOZ3yRnwrGasmA5BSQnJYGcpAHgQCYnIAQM2XJ-aCDhSAZ+xwwkqBqecIlASEVwCqFUIAA)

src/generator.ts
bcherny: That's embarrassing, good catch :) Would `if (ast.spreadParam) {` work?

bcherny:
Published 7.0.0.

PR #242, opened 2019-07-02
feat: support `minItems` and `maxItems`

### LINKED ISSUE #218: Tuple items with `minItems` and `additionalItems`

bradzacher:
NOTE - this branch was based off of #241.
Github will show a number of unrelated commits until that PR is merged.
For a nicer reviewing experience, please [only review commits after `5cb467a`](https://github.com/bcherny/json-schema-to-typescript/pull/242/files/2eb3a9f4cccbea1b8a9281bad10dacf0946666f5..ffb52f2f011bad563f355521b492aaac2c024ece)

---

Fixes #218

This PR:
- add support for generating tuple types with optional and non-optional elements if `minItems` / `maxItems` exists.
- handles having either `minItems` / `maxItems` greater than the length of the provided items.
- handles having `maxItems` less than the length of the provided items.
- converts array types to tuple types in the presence of `minItems` or `maxItems`.


src/generator.ts
```diff
@@ -158,9 +170,65 @@ function generateType(ast: AST, options: Options): string {
...
+          }
+        } else {
+          // no max items and no spread param, so just spread any
+          spreadParam = T_ANY
+        }
+      }
+      if (maxItems > astParams.length && ast.spreadParam === undefined) {
+        // this is a valid state, and JSONSchema doesn't care about the item type
+        // fill the tuple with any elements
+        for (let i = astParams.length; i < maxItems; i += 1) {
```
bcherny: Is there any way to simplify this logic? It feels a little bit repetitive.

src/parser.ts
```diff
@@ -69,6 +69,37 @@ function parseNonLiteral(
 
   log(whiteBright.bgBlue('parser'), schema, '<-' + typeOfSchema(schema), processed.has(schema) ? '(FROM CACHE)' : '')
 
+  function parseArray(
+    standaloneName: string | undefined,
+    params: AST,
+    minItems: number | undefined,
+    maxItems: number | undefined
+  ): AST {
+    if ((typeof minItems === 'number' && minItems > 0) ||
```
bcherny: Is this something that you could move to the normalizer step, so that you unit test it in isolation?

src/generator.ts
```diff
@@ -158,9 +170,65 @@ function generateType(ast: AST, options: Options): string {
...
+          astParams.push(T_ANY)
+        }
+      }
+
+      const params = astParams
+        .map(param => generateType(param, options))
+        .map((param, index) => {
+          if (maxItems > 0 && index >= maxItems) {
+            // it's perfectly valid to provide 5 item defs but require maxItems 1
```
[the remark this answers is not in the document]
bradzacher: I'm thinking not, just because that would mean the normalizer would have to do handling of the `oneOf`/`anyOf` resolution.

bradzacher:
I've addressed the other comments.

In regards to changing the types, I think it's a good idea.
I'll have a look at it tonight.

bradzacher:
It makes the types verbose, but it adds a heap more type safety.
It also allows explicit assertion of the empty array state when no `minItems` is provided.

src/generator.ts
```diff
@@ -158,9 +170,83 @@ function generateType(ast: AST, options: Options): string {
     case 'OBJECT': return 'object'
     case 'REFERENCE': return ast.params
     case 'STRING': return 'string'
-    case 'TUPLE': return '['
-      + ast.params.map(_ => generateType(_, options)).join(', ')
-      + ']'
+    case 'TUPLE': return (() => {
+      const minItems = toNumber(ast.minItems)
+      const maxItems = toNumber(ast.maxItems)
```
bcherny: Should this defaulting go in a normalizer step?

src/parser.ts
```diff
@@ -196,10 +207,28 @@ function parseNonLiteral(
     case 'UNNAMED_SCHEMA':
       return set(newInterface(schema as SchemaSchema, options, rootSchema, processed, usedNames, keyName, keyNameFromDefinition))
     case 'UNTYPED_ARRAY':
+      const {minItems, maxItems} = schema
+      const params = T_ANY
+      if ((typeof minItems === 'number' && minItems > 0) ||
+        (typeof maxItems === 'number' && maxItems > 0)) {
```
bcherny: If you accept my previous suggestion of adding another normalizer step, it could simplify this logic.

test/testNormalizer.ts
```diff
@@ -20,9 +20,11 @@ export function run() {
     .map(_ => [_, require(_)] as [string, JSONTestCase])
     .forEach(([filename, json]: [string, JSONTestCase]) => {
       const params = { filename }
-      test(json.name, t =>
-        t.snapshot(template(toString(normalize(json.in, filename)))(params))
-      )
+      test(json.name, t => {
+        const normalised = normalize(json.in, filename);
+        t.snapshot(template(toString(normalised))(params))
+        t.deepEqual(json.out, normalised);
```
bcherny: Good catch!

bradzacher:
OOOOKay. Sorry, this PR keeps growing.

So in adding in a normalizer for `minItems`, I noticed that the normalizer wasn't traversing correctly. I added a naive normaliser (`if (!('minitems' in schema)) { schema.minItems = 0 }`), and it broke most of the tests because the normalization algorithm traversed every single object key equally.

I noticed that most of the normalizers had a check for `stringify(schema) === stringify(rootSchema)`, i.e. only apply at the root, but this means that they don't get applied correctly for anything beyond that - i.e. definitions, allOf, anyOf, oneOf, not, items, properties....

So I added a traverser which will properly traverse the schema, which allows the normalizers to correctly apply to things which are actually schemas beyond the root schema.

This surfaced a few bugs which I fixed.

bcherny:
Published 7.0.0.

PR #245, opened 2019-07-12
Fix TS properties are optional for compile function

egmen:
In README.MD is written that all properties are optional.
This is a fix of types


bcherny:
Thanks for the contribution, but this isn't necessary, since we make the `Options` object have all-optional members here already: https://github.com/bcherny/json-schema-to-typescript/blob/fcbe243e62df5722cda1cffa4049c3942d519f8b/src/index.ts#L74

PR #261, opened 2019-09-30
Fix `UNION` and `NULL` types

gmathieu:
This fixes the use of UNION and NULL types (ex. `{ type: ['null', 'string'] }`).

See snapshot diff to get a clear idea of what's happening

This is a 3 part fix:
1. missing schema when mapping over UNION type: `{ type: _ }` vs `{ ...schema, type: _ }`
1. prioritize UNION type in `typeOfSchema`
1. prioritize NULL type in `typeOfSchema`

The parser needs to process `UNION` and `NULL` before other types suchs as `ENUM`, `ARRAY`...


gmathieu:
@bcherny any thoughts?

gmathieu:
Hi, does anyone have any feedback?

bcherny:
Published `8.0.0`.

PR #264, opened 2019-10-21
Convert json schema draft 06 const values to typescript const values

### LINKED ISSUE #263: Support for constant values
Support const values (new in v6). https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values

Input

```json
{
  "$id": "https://api.nn-group.com/t/nn-nl/2/json_schemas/forwardcache_notify_participation",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "Foo",
  "required": ["bar"],
  "properties": {
    "bar": {
      "const": true
    }
  }
}
```

Expected output:

```ts
export interface Foo {
  bar: true;
}
```

Current output:

```ts
export interface Foo {
  bar: {
    [k: string]: any;
  };
}
```

This feature was mentioned in #120 but as the scope of that is a bit too big for beginning contributors I thought I would split this feature out.

Any hints for json-schema-to-typescript newbies would be much appreciated.

djbeaumont:
Resolves #263

I'm a little unclear on the structure of the project so any feedback on the PR is gratefully received. This is a little exploratory and I have no idea if this can be said to implement the feature from json schema draft 06, but I think it covers enough for my needs.


nicojs:
Awesome! Open source at it's finest 👍. I like the simplicity of the solution.

The simple example now works:

```ts
export interface Foo {
  bar: true;
}
```

However, it seems that falsy values don't work yet 😈.

```json
{
  "title": "Foo",
  "required": ["bar", "baz", "qux", "quux"],
  "type": "object",
  "properties": {
    "bar": {
      "const": false
    },
    "baz": {
      "const": ""
    },
    "qux": {
      "const": 0
    },
    "quux": {
      "const": null
    }
  }
}

```

This results in:

```ts
export interface Foo {
  bar: {
    [k: string]: any;
  };
  baz: {
    [k: string]: any;
  };
  qux: {
    [k: string]: any;
  };
  quux: {
    [k: string]: any;
  };
}
```

nicojs:
> I'll see your PR and I'll raise you another PR 

The falsy value thingy should be fixed in https://github.com/djbeaumont/json-schema-to-typescript/pull/1

🤟

djbeaumont:
ahh great catch, thank you!

djbeaumont:
@bcherny what do you think of this? I suppose it would have to be a new major version as it will be a big change to the outputted types.

nicojs:
@bcherny I don't want to be rude, but could you please take a look? I would like to actually use it at our company. Thanks!

qzb:
@nicojs Unfortunately bcherny maintains this package pretty unfrequently, I guess he'll accept this PR, but it could take a while.

djbeaumont:
@nicojs FWIW I'm using this workaround in my codebase:
```
/**
 * Workaround until this is merged:
 * https://github.com/bcherny/json-schema-to-typescript/pull/264
 * @param {string} definition
 * @returns {string} Same defined interface with event_type made constant.
 */
function fudgeEventTypeInDefinition(eventType, definition) {
  const toReplace = 'event_type: string;';
  const replaceWith = `event_type: '${eventType}';`;
  return definition.replace(toReplace, replaceWith);
}
```
It works because the constant property I want to set happens to be the same as the filename of the schema that I'm parsing. Clearly not a general solution.

qzb:
@djbeaumont  According to JSON Schema const is only a shorthand for enum with single element, so if you can modify original JSON Schema file, simply replace `"const": "value"` with `"enum": [ "value" ]`. Enums are handled correctly by json2ts, so it should workaround your issue.

JonathanWilbur:
@djbeaumont I can't wait for this to make it in. Thank you so much for pursuing this! Do you expect to follow up soon with corrections so this can be merged in?

jeffvandyke:
What's the status on this? I found myself wanting to use `"const"`, and while single-member `"enum"` works, I'd be excited to see this make it in!

L2jLiga:
Hi there? Any updates upcomming here?

bcherny:
https://github.com/bcherny/json-schema-to-typescript/pull/290 is another shot at this, using the normalizer approach.

PR #274, opened 2020-01-23
Added ignoreMinAndMaxItems option

### LINKED ISSUE #250: Add switch to disable the minItems/maxItems expanding
In 0.7.0 a new PR https://github.com/bcherny/json-schema-to-typescript/pull/242 is merged to support minItems/maxItems for the array, this is great!

Until something like
```
        "processes": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Process"
          },
          "minItems": 1,
          "maxItems": 512,
```
In my JSON Schema...

This kind of schema generate a HUGE interface file which is not usable. May I ask can you add a flag to disable this feature or do some limitation like convert only for the case less than 5 items? Thanks.

lukeggchapman:
Fixes #250 

Added ignoreMinAndMaxItems to prevent tuples from being generated for array types.

Tuples can cause issues when there is a high value for maxItems, and they don't work well with array utility methods (.map, .filter). It's unlikely that TS will fix these issues anytime soon given tuples should be a fixed length.


src/normalizer.ts
```diff
@@ -4,7 +4,11 @@ import {cloneDeep} from 'lodash'
 import {JSONSchema, JSONSchemaTypeName, NormalizedJSONSchema} from './types/JSONSchema'
 import {escapeBlockComment, justName, log, toSafeString, traverse} from './utils'
 
-type Rule = (schema: JSONSchema, rootSchema: JSONSchema, fileName?: string) => void
+export interface NormalizeOptions {
```
bcherny: Just use the existing `Options` type here -- no need to add another type.

src/normalizer.ts
```diff
@@ -4,7 +4,11 @@ import {cloneDeep} from 'lodash'
...
+  ignoreMinAndMaxItems: boolean
+}
+
+type Rule = (schema: JSONSchema, rootSchema: JSONSchema, fileName?: string, options?: NormalizeOptions) => void
```
bcherny: To simplify line 71, can you make `options` required?

src/normalizer.ts
```diff
@@ -63,7 +67,21 @@ rules.set('Escape closing JSDoc Comment', schema => {
   escapeBlockComment(schema)
 })
 
-rules.set('Normalise schema.minItems', schema => {
+rules.set('Optionally remove maxItems and minItems', (schema, _rootSchema, _fileName, options) => {
+  if (options && options.ignoreMinAndMaxItems) {
+    if (schema.maxItems !== undefined) {
```
bcherny: Ignorable suggestion: Maybe cleaner to use an `in` or `hasOwnProperty` check?

src/normalizer.ts
```diff
@@ -72,11 +90,15 @@ rules.set('Normalise schema.minItems', schema => {
   // cannot normalise maxItems because maxItems = 0 has an actual meaning
 })
 
-rules.set('Normalize schema.items', schema => {
+rules.set('Normalize schema.items', (schema, _rootSchema, _fileName, options) => {
   const {maxItems, minItems} = schema
   const hasMaxItems = typeof maxItems === 'number' && maxItems >= 0
   const hasMinItems = typeof minItems === 'number' && minItems > 0
 
+  if (options && options.ignoreMinAndMaxItems) {
```
bcherny: Mind moving this to the top of this function?

bcherny:
Published 8.1.0.

Issue #279, opened 2020-03-04
Allow unicode characters in type name


zhangciwu:
https://github.com/bcherny/json-schema-to-typescript/blob/d898361ff973d3746ea695f1efce75c975061c6f/src/utils.ts#L164

The member of type name is restricted to only english letters (and some symbol)
Typescript now allows unicode characters ([doc link](https://www.typescriptlang.org/docs/handbook/release-notes/overview.html#better-unicode-support-for-identifiers))

So, modify this function to allow unicode characters in type name, or as an option?

zhangciwu:
My scenario: `title` value is in full chinese, so the result typename is blank
(Maybe another bug to fix: do not generate blank type name)

zhangciwu:
A minimium schema would cause this error:
```json
    {
            "type": "object",
            "title": "呵呵",
            "properties": {
                "some_key": {
                    "type": "string",
                    "title": "哈哈"
                }
            }
        }
```

bcherny:
Sounds good to me! Want to put up a Pull Request?

zhangciwu:
Fix the blank name bug in #280 

Allow unicode is a little complicated, would be later

bcherny:
Thanks for the fix, but I think supporting Unicode (as you suggested) could be a better way to do it. What issues are you running into?

zhangciwu:
Got a doc: [Valid JavaScript variable names in ES2015](https://mathiasbynens.be/notes/javascript-identifiers-es6)
Maybe using this code provided in this page to judge a string is a valid identifier is comprehensive, but it's not packed into npm, and while I found some packs:

https://github.com/purplecabbage/valid-identifier
https://github.com/YerkoPalma/is-valid-identifier
https://github.com/SteveWestbrook/is-valid-var-name

First download for most times, but seems not active; others are not widely adopted (npm downloads)
A little hard to choose

iansw246:
I am working on this.

I wanted to confirm that this project targets ES5. I'm considering using Unicode aware regex, but it appears to be only supported from ES2015.

If using ES5, I found https://github.com/mathiasbynens/regexpu.

Additionally, are there any concerns about backwards compatibility or previously generated names?

PR #280, opened 2020-03-10
fix: blank name if all characters filter out

zhangciwu:
If all characters filter out at `toSafeString`, the output would be a blank name, thus cause an error when examine the result. (example in #279 )

This fix is simple, when safe name is blank, set it to `NoName`

Test added.


zhangciwu:
Merged.
(test.ts.snap regenerated, this file seems can be ignore)

bcherny:
Thanks for the contribution! Will publish as part of 9.0.0 this week.

Issue #259, opened 2019-09-22
Cannot use 'in' operator to search for 'required' in true


felixfbecker:
Getting this error when upgrading to 7.1:
```

TypeError: Cannot use 'in' operator to search for 'required' in true
--
  | [2019-09-09T06:16:07Z]     at /buildkite/builds/buildkite-agent-7b9d66c86c-qjbr5-1/sourcegraph/.golang/sourcegraph/src/github.com/sourcegraph/sourcegraph/node_modules/json-schema-to-typescript/dist/src/normalizer.js:23:22
  | [2019-09-09T06:16:07Z]     at /buildkite/builds/buildkite-agent-7b9d66c86c-qjbr5-1/sourcegraph/.golang/sourcegraph/src/github.com/sourcegraph/sourcegraph/node_modules/json-schema-to-typescript/dist/src/normalizer.js:80:62
  | [2019-09-09T06:16:07Z]     at traverse (/buildkite/builds/buildkite-agent-7b9d66c86c-qjbr5-1/sourcegraph/.golang/sourcegraph/src/github.com/sourcegraph/sourcegraph/node_modules/json-schema-to-typescript/dist/src/utils.js:92:5)
  | [2019-09-09T06:16:07Z]     at traverse (/buildkite/builds/buildkite-agent-7b9d66c86c-qjbr5-1/sourcegraph/.golang/sourcegraph/src/github.com/sourcegraph/sourcegraph/node_modules/json-schema-to-typescript/dist/src/utils.js:117:13)
  | [2019-09-09T06:16:07Z]     at /buildkite/builds/buildkite-agent-7b9d66c86c-qjbr5-1/sourcegraph/.golang/sourcegraph/src/github.com/sourcegraph/sourcegraph/node_modules/json-schema-to-typescript/dist/src/utils.js:84:13
  | [2019-09-09T06:16:07Z]     at Array.forEach (<anonymous>)
  | [2019-09-09T06:16:07Z]     at traverseObjectKeys (/buildkite/builds/buildkite-agent-7b9d66c86c-qjbr5-1/sourcegraph/.golang/sourcegraph/src/github.com/sourcegraph/sourcegraph/node_modules/json-schema-to-typescript/dist/src/utils.js:82:22)
[...]

PR #284, opened 2020-03-19
Fix cannot use 'in operator  #259

LutzR:


bcherny:
Thanks for the contribution! Would you mind adding a test for this, so we can make sure it doesn't regress in the future?

ohana54:
@bcherny I took the liberty to build on the great work done here and added a test: #310 
I hope that's ok. @LutzR if you're available, feel free to take the test and add it here, and I will close my PR in favor of this one.

bcherny:
https://github.com/bcherny/json-schema-to-typescript/pull/310

PR #281, opened 2020-03-14
Adding ability to convert any's into unknown's

dl748:
closes #275

PR #291, opened 2020-04-21
Add strictUnknownTypes option

mike-marcacci:
When set to true, this option will type unknown values as `unknown` instead of `any`, requiring further type refinement to use.


mike-marcacci:
As an additional note, these tests fail nondeterministically on master, so their potential presence in CI is expected and not related to this PR:

```
error Error at key "bar" in file "enumValidation.1": Enum members and tsEnumNames must be of the same length
error Error at key "bar" in file "enumValidation.2": tsEnumNames must be an array of strings
```

bcherny:
Thanks for the contribution, but this is a dup of #281

PR #295, opened 2020-05-04
bug fixes in cli

Liooo:
Thanks for the great work, found some bug like behaviors in cli so please review.

- had redundant directroy in output like `*/data.d.ts/data.d.ts`
- normalize path to avoid error when complicated path is specified
- any directory names included in the input path was removed from output
- wrong dir structure when input and output paths have no intersection


src/utils.ts
```diff
@@ -238,14 +238,10 @@ And the user wants the outputs to be in MultiSchema/Out, then this code will be
   MultiSchema/Out/bar/fuzz/c.json
   MultiSchema/Out/bar/d.json
 */
-export function pathTransform(o: string, i: string): string {
-  const outPathList = o.split('/')
-  const inPathList = i.split('/')
+export function pathTransform(o: string, i: string, f: string): string {
```
bcherny, line 241: Would you mind improving the naming for these params, while you’re at it?

Liooo:
@bcherny ptal 6a88f5b 🙏

bcherny:
Published 9.1.0.

PR #212, opened 2019-01-08
Interfaces now extend custom types

shripadk:
Fixes #211

PR #302, opened 2020-05-18
Interfaces now extend custom types

tivervac:
Replaces https://github.com/bcherny/json-schema-to-typescript/pull/212


tivervac:
Since work on #212 appears to have stopped, I forked the commit rebased it and added a new test for the other test case in the thread.

tivervac:
The `CONTRIBUTING.md` mentions that we should use `node-inspector` for debugging but it seems that package is deprecated, any new instructions on how to debug a test? Specifically I want to try and check on the comments you left on #212

src/generator.ts
```diff
@@ -343,7 +343,7 @@ function generateStandaloneInterface(ast: TNamedInterface, options: Options): st
     (hasComment(ast) ? generateComment(ast.comment) + '\n' : '') +
     `export interface ${toSafeString(ast.standaloneName)} ` +
     (ast.superTypes.length > 0
-      ? `extends ${ast.superTypes.map(superType => toSafeString(superType.standaloneName)).join(', ')} `
+      ? `extends ${ast.superTypes.map(superType => toSafeString(superType.standaloneName || '')).join(', ')} `
```
tivervac, line 346: Default needed here because `CustomType` doesn't have the `standaloneName` property thus by making `ast.superTypes` return an array of `TNamedInterface | TCustomType` standaloneName can now be undefined as well

src/generator.ts
```diff
@@ -338,12 +338,17 @@ function generateStandaloneEnum(ast: TEnum, options: Options): string {
   )
 }
 
+function notUndefined<T>(x: T | undefined): x is T {
+  return x !== undefined
+}
+
 function generateStandaloneInterface(ast: TNamedInterface, options: Options): string {
+  const superTypeNames = ast.superTypes.map(x => x.standaloneName).filter(notUndefined)
```
tivervac, line 346: This is needed because `ast.superTypes` now returns `TNamedInterface | TCustomType` and `TCustomType` doesn't override `standaloneName` to be a required property like `TNamedInterface` does.

I don't see `TCustomType` used anywhere except in this feature. I could thus also make `standaloneName` required in `TCustomType` or I could create a named `TCustomType` that way errors can be caught at parse time instead of at generation time

src/parser.ts
```diff
@@ -306,17 +308,44 @@ function parseSuperTypes(
...
+  if (schema.tsType) {
+    return {
+      comment: schema.description,
+      keyName: schema.title,
+      params: schema.tsType,
+      standaloneName: standaloneName(schema, schema.title, usedNames),
+      type: 'CUSTOM_TYPE'
+    }
+  } else {
+    throw Error(format('Customtype must have a tsType', schema))
```
tivervac, line 347: This method is only called in 1 place where `schema.tsType` is already checked so we should be good, this is mostly a guard against programming mistakes.

CONTRIBUTING.md
```diff
@@ -11,3 +11,6 @@
 - To debug a test, with breakpoints:
   1. Install [node-inspector](https://www.npmjs.com/package/node-inspector)
   2. Run `node-debug ./node_modules/ava/profile.js ./dist/test/test.js`
+- When updating the e2e tests:
+  1. Run `yarn test`
+  2. If the snapshot changed and the changes are as expected: run `ava --update-snapshots`
```
bcherny, line 16: 👍

src/parser.ts
```diff
@@ -306,17 +308,59 @@ function parseSuperTypes(
...
+    const extendedType = newExtendedType(schema, usedNames)
+    processed.set(schema, extendedType)
+    return extendedType
+  }
+  const namedInterface = newNamedInterface(schema, options, rootSchema, processed, usedNames)
+  processed.set(schema, namedInterface)
+  return namedInterface
+}
+
+function newExtendedType(schema: SchemaSchema, usedNames: UsedNames): TCustomType {
```
bcherny, line 352: Can we tighten up the types here?

Then, you can remove the if/else/throw below.

```suggestion
function newExtendedType(schema: CustomTypeJSONSchema, usedNames: UsedNames): TCustomType {
```

src/parser.ts
```diff
@@ -306,17 +308,59 @@ function parseSuperTypes(
...
+}
+
+function getOrCreateSuperType(
+  schema: SchemaSchema,
+  options: Options,
+  rootSchema: JSONSchema,
+  processed: Processed,
+  usedNames: UsedNames
+): TSuperType {
+  if (processed.has(schema)) {
```
bcherny, line 335: What sort of cycle is this catching? Mind adding a test case for it?

src/parser.ts
```diff
@@ -306,17 +308,59 @@ function parseSuperTypes(
...
+    const hit = processed.get(schema)
+    if (isSuperType(hit)) {
```
bcherny, line 337: This check is redundant, right?

src/parser.ts
```diff
@@ -306,17 +308,59 @@ function parseSuperTypes(
...
+function newExtendedType(schema: SchemaSchema, usedNames: UsedNames): TCustomType {
+  if (schema.tsType) {
+    return {
+      comment: schema.description,
+      keyName: schema.title,
+      params: schema.tsType,
+      standaloneName: standaloneName(schema, schema.title, usedNames),
+      type: 'CUSTOM_TYPE'
+    }
+  } else {
```
bcherny, line 361: Mind throwing an error & copying the TODO from `newNamedInterface` here, so that we enforce that this always has a standalone name early, and avoid having to handle this case downstream?

Or, compute a `standaloneName` here if one isn't defined.

src/types/AST.ts
```diff
@@ -6,6 +6,8 @@ export type AST = TAny | TArray | TBoolean | TEnum | TInterface | TNamedInterfac
   | TIntersection | TLiteral | TNumber | TNull | TObject | TReference
   | TString | TTuple | TUnion | TCustomType
 
+export type TSuperType = TNamedInterface | TCustomType
```
bcherny, line 9: Could there be a way to indicate with the types that `standaloneName` is always defined for supertypes? eg.
```suggestion
export type TSuperType = ASTWithStandaloneName & (TNamedInterface | TCustomType)
```

src/generator.ts
bcherny: Let's see if we can explicitly handle this case upstream (see my previous suggestion), instead of needing to implicitly handle it here.

src/parser.ts
tivervac: This just protects against programming mistakes. At this point schema should've been processed already. If it hasn't, something went wrong earlier. Guess I could throw an error in that case

src/parser.ts
tivervac: Depends, do you want the code to be robust against a broken schema? I could write a (wrong) schema that extends something that's not a supertype

stephanrotolante:
can we merge this

src/parser.ts
tivervac: Alright

src/parser.ts
tivervac: I've added a comment

src/parser.ts
tivervac: Done

src/types/AST.ts
tivervac: Done

src/parser.ts
tivervac: Added the error

tivervac: The function is a get or create. If processed has schema, it's a get, otherwise it's a create. The tests already cover both branches

tivervac:
@stephanrotolante We can once @bcherny approves my changes

stephanrotolante:
@tivervac I know, he's been M.I.A for quite some time now

src/parser.ts
```diff
@@ -305,17 +306,63 @@ function parseSuperTypes(
...
     return []
   }
   if (Array.isArray(superTypes)) {
-    return superTypes.map(_ => newNamedInterface(_, options, _, processed, usedNames))
+    return superTypes.map(_ => getOrCreateSuperType(_, options, _, processed, usedNames))
+  }
+  return [getOrCreateSuperType(superTypes, options, superTypes, processed, usedNames)]
+}
+
+function isSuperType(type: any): type is TSuperType {
```
bcherny, line 322: Mind moving this to AST.ts?

Also, let's tighten up the types here if possible:

```suggestion
function isSuperType(type: AST): type is TSuperType {
```

src/parser.ts
bcherny: Validating the schema seems reasonable to me. Mind moving this check to validator.ts, to decouple it from the parsing stage?

src/parser.ts
```diff
@@ -305,17 +306,63 @@ function parseSuperTypes(
...
+    // Protect against programming mistakes
+    if (isSuperType(hit)) {
+      return hit
+    } else {
+      throw Error(format('type is not a supertype', hit))
+    }
+  }
+
+  if (schema.tsType) {
+    const extendedType = newExtendedType(schema as CustomTypeJSONSchema, usedNames)
```
bcherny, line 344: Mind creating a new util for this, to avoid the type assertion?

```ts
// JSONSchema.ts
export function isCustomType(schema: LinkedJSONSchema): ast is CustomTypeJSONSchema {
  return 'tsType' in schema
}
```

src/parser.ts
```diff
@@ -305,17 +306,63 @@ function parseSuperTypes(
...
+    processed.set(schema, extendedType)
+    return extendedType
+  }
+  const namedInterface = newNamedInterface(schema, options, rootSchema, processed, usedNames)
+  processed.set(schema, namedInterface)
+  return namedInterface
```
bcherny, line 350: Nit: Can you avoid the duplication here?

```ts
const supertype = isCustomType(schema)
  ? newExtendedType(schema, usedNames)
  : newNamedInterface(schema, options, rootSchema, processed, usedNames)
processed.set(schema, supertype)
return supertype
```

src/parser.ts
```diff
@@ -305,17 +306,63 @@ function parseSuperTypes(
...
+}
+
+function newExtendedType(schema: CustomTypeJSONSchema, usedNames: UsedNames): TSuperType {
+  const name = standaloneName(schema, schema.title, usedNames)
+  if (!name || name.length == 0) {
```
bcherny, line 355: This is redundant.
```suggestion
  if (!name) {
```

src/parser.ts
```diff
@@ -305,17 +306,63 @@ function parseSuperTypes(
...
+    return superTypes.map(_ => getOrCreateSuperType(_, options, _, processed, usedNames))
+  }
+  return [getOrCreateSuperType(superTypes, options, superTypes, processed, usedNames)]
+}
+
+function isSuperType(type: any): type is TSuperType {
+  return type.type === 'CUSTOM_TYPE' || type.type === 'INTERFACE'
+}
+
+function getOrCreateSuperType(
```
bcherny, line 326: Nit: please rename this for consistency
```suggestion
function newSupertype(
```

src/parser.ts
```diff
@@ -305,17 +306,63 @@ function parseSuperTypes(
...
+    const extendedType = newExtendedType(schema as CustomTypeJSONSchema, usedNames)
+    processed.set(schema, extendedType)
+    return extendedType
+  }
+  const namedInterface = newNamedInterface(schema, options, rootSchema, processed, usedNames)
+  processed.set(schema, namedInterface)
+  return namedInterface
+}
+
+function newExtendedType(schema: CustomTypeJSONSchema, usedNames: UsedNames): TSuperType {
```
bcherny, line 353: Do you need this function, or can you call `parse` directly?

ie. in `newSupertype`:

```ts
const supertype = isCustomType(schema)
  ? parse(schema, options, keyName, processed, usedNames)
  : newNamedInterface(schema, options, rootSchema, processed, usedNames)
processed.set(schema, supertype)
return supertype
```

bcherny (changes requested):
Sorry for the delay!

This PR is looking good. Let's take one more pass to address a few minor comments. Also, please rebase.

src/parser.ts
tivervac: Done

src/parser.ts
tivervac: Done

src/parser.ts
tivervac: I have no idea

tivervac:
Sadly it seems like the API on master has changed quite a bit since the long time I wrote this. I have since long outgrown the need for this patch, or even library, and I have no more real life test case to test this again. Even just rebasing against master or answering some slightly deeper questions now seem like a burden. It doesn't look like I have the will, time or expertise to pick this up again. If anyone else is interested, feel free to use the code

bcherny:

PR #305, opened 2020-05-22
Fix empty interfaces for multi patternProperties

ggoodman:
When a schema declared multiple `patternProperties` and no `additionalProperties`, the resulting type definition was effectively 'empty'.

This PR produces something that makes a bit more intuitive sense to me for when a schema has no `additionalProperties` and `patternProperties`.


sebas2day:
I believe this addition would make it more complete. As an example, try compiling the [OpenAPI3 json schema spec](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v3.0/schema.json) into TypeScript definitions. In order to make it usable I need to extend the Paths interface since its empty by default using this tool. Ideally you would like it to function out-of-the-box. However, I'm not sure whether the behavior of this PR is correct since I get an Unknown (with capital) when compiling the OpenAPI. It might need some more fine tuning

bcherny:

PR #312, opened 2020-06-12
make types nullable

TimHaboldt:
Hey.
Was using your typescript compiler for a while and noticed there is no support for the nullable flag. I tried to implement a solution. I hope it satisfies you enough to make it into the main repository. Would be pleased to use this feature in future.
Ps. I implemented a solution that works for me. I cannot promise every possible nullable flag will work with those changed. The most important ones should be covered.
Supported are all basic types and unnamed enums. Named enums, objects and Interfaces shouldn`t be covered.

Regards

### DIFF
--- a/src/generator.ts
+++ b/src/generator.ts
@@ -186,7 +186,7 @@ function generateRawType(ast: AST, options: Options): string {
         return type.endsWith('"') ? '(' + type + ')[]' : type + '[]'
       })()
     case 'BOOLEAN':
-      return 'boolean'
+      return 'boolean' + (ast.isNullable? ' | null' : '')
     case 'INTERFACE':
       return generateInterface(ast, options)
     case 'INTERSECTION':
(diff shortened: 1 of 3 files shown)


bcherny:
Hey there, thanks for the contribution! I don't see `nullable` in the JSON-Schema spec -- mind linking to it in your summary?

TimHaboldt:
> Hey there, thanks for the contribution! I don't see `nullable` in the JSON-Schema spec -- mind linking to it in your summary?

I looked into it. It was a fault on my part.
It is a specification of the open API Schema (https://swagger.io/specification/) which we are currently using.
Therefore it opens the question if you would like to support the open API Schema too.
I would realize an open API flag you can set inside the options that would allow the open API Schema support.

zmeyc:
@bcherny We also encountered this issue with nullable types. What do you think about supporting OpenAPI, are any changes needed for this PR or is it out of scope at all?

bcherny:
Let's support this, but I'd suggest a slightly different implementation: instead of handling `isNullable` in the parser, generator, and IR, can you instead add a normalizer step that converts `isNullable` annotations to unions with null (ie. add it to the `type` array, and make `type` an array if it wasn't already)?

```ts
// normalizer.ts
rules.set('OpenAPI: Convert isNullable to union with null', schema => {
  // ...
}
```

This PR also needs a unit test for your new normalizer rule, as well as an e2e test.

bcherny (changes requested):
Thanks for the contribution! Let's take one more pass at this.

zone117x:
Looks like `nullable` is also now supported in redoc@2.x. Would be great to have it supported here. 
@bcherny I haven't looked into the code for this lib -- do you think implementing the changes you last suggested would be easy for a new contributor?

Zusendo:
@zone117x I would be grateful if someone else would implement the changes for me. My current workload does not support any other contributions right now.

TimHaboldt:
@zone117x  I appreciate if someone else could take over. My workload does not allow me to work at anything else right now.

zone117x:
I might be able to work on it if @bcherny is available for reviewing the PR and giving guidance where needed (for example how the tests should be structured, etc)

bcherny:
Superceded by https://github.com/bcherny/json-schema-to-typescript/pull/411

PR #313, opened 2020-06-22
Support schema module imports

ruswerner:
I added support for a `tsModule` property. When specified, this adds an `import` declaration at the top of the generated file. We needed this feature so that the exported types that reference other schemas could be imported and not inline duplicated. My main motivation was to unblock our team and reduce error-prone manual steps when generating types. I'm providing this PR to open some discussion and solicit ideas before taking it any further (as there are multiple ways to solve this problem).

Usage:

    ...
    someColourField: {$ref: 'https://myorg.com/schemas/colours.json', tsModule: './colours', tsType: 'Colour' },
    ...

will produce:

    import {Colour} from './colours';
    ...
    export interface MyObject {
        ...
        someColourField: Colour;
        ...
    }

Import statements are de-duped, however it doesn't support importing multiple types from the same module; we didn't have this need and generally only reference the single top-level schema object exported from a given module.

All feedback welcome!

### DIFF
--- a/src/generator.ts
+++ b/src/generator.ts
@@ -1,5 +1,5 @@
 import {whiteBright} from 'cli-color'
-import {omit} from 'lodash'
+import {omit, uniq} from 'lodash'
 import {DEFAULT_OPTIONS, Options} from './index'
 import {
   AST,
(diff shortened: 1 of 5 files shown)


ruswerner:
@bcherny I'm going to close this PR as we went in a different direction and I tend to agree with your assessment. Thanks for the work on this repo!

Elvennn:
@bcherny @ruswerner I would really need something like this. Generating imports between different generated schemas instead of duplicating every reference in each file.

My use case: I have a model definition in JSON-Schema (~70 types) referencing each others. I want to generate TypeScript interfaces for everything, with a type per file with all of its generated referenced types imported.

The kind of mechanism describe in this PR is a solution

PR #318, opened 2020-07-04
Handle internal refs so they don't duplicate with different descriptions

### LINKED ISSUE #193: Duplication of type : properties with different description
Hi there !

I have a problem with property, ref and description. If two properties of a type have the same ref definitions and two different descriptions, `json2ts` duplicate the type definition.

In the following case, I have a type named `Offer` containing two properties `price` and `priceInclVAT` of the same type `Price` but two different descriptions "Price excl. VAT" and "Price incl. VAT". Output from `json2ts` creates two types `Price` and `Price1` and put comments over interface definitions instead of property definitions.

Source JSON Schema :
```json
{
    "title": "Offer",
    "description": "An offer",
    "type": "object",
    "properties": {
        "price": {
            "description": "Price excl. VAT",
            "$ref": "#/definitions/Price"
        },
        "priceInclVAT": {
            "description": "Price incl. VAT",
            "$ref": "#/definitions/Price"
        }
    },
    "definitions": {
        "Price": {
            "title": "Price",
            "description": "A price",
            "type": "object",
            "properties": {
                "value": {
                    "description": "Price as number",
                    "type": "number"
                },
                "text": {
                    "description": "Price as string",
                    "type": "string"
                }
            }
        }
    }
}
```

What I obtained using  `json2ts` :
```ts
/**
 * An offer
 */
export interface Offer {
    price: Price;
    priceInclVAT: Price1;
}

/**
 * Price excl. VAT
 */
export interface Price {
    /**
     * Price as number
     */
    value: number;
    /**
     * Price as string
     */
    text: string;
}

/**
 * Price incl. VAT
 */
export interface Price1 {
    /**
     * Price as number
     */
    value: number;
    /**
     * Price as string
     */
    text: string;
}
```

What I expected : 
```ts
/**
 * An offer
 */
export interface Offer {
    /**
     * Price excl. VAT
     */
    price: Price;
    /**
     * Price incl. VAT
     */
    priceInclVAT: Price;
}
/**
 * A price
 */
export interface Price {
    /**
     * Price as number
     */
    value: number;
    /**
     * Price as string
     */
    text: string;
}
```

davedoesdev:
Fixes #193 with new `resolve` option set to `false`


davedoesdev:
I tried an approach which removed the `description` fields and then did a deep comparison. That does work but resulted in referencing interfaces that may be the same in structure but weren't originally referenced in the schema.

The JSON-Schema-Ref-Parser should (somehow) be maintaining the same object equality but set a different attribute.

I don't have a lot of time to work on this PR, I'm just using the quick fix to make my output sane. Feel free to close it if you like.

bcherny:
> I tried an approach which removed the description fields and then did a deep comparison. That does work but resulted in referencing interfaces that may be the same in structure but weren't originally referenced in the schema.

I think you can use a shallow comparison, right? Per (1) and (2) above, we should maintain referential identity, even after de-referencing, and if we don't, that should be fixed as part of this PR (or, as a pre-req for it).

davedoesdev:
> I think you can use a shallow comparison, right? Per (1) and (2) above, we should maintain referential identity, even after de-referencing, and if we don't, that should be fixed as part of this PR (or, as a pre-req for it).

Yes, I agree. Sorry, I just don't have time to work on this more. Perhaps someone else can help out?

ttyniwa:
@bcherny 
Thank you for your great library.
I'm afraid this is a layman's question, but wouldn't it be an easy solution if we used bundle instead of json-schema-ref-perser dereference?

bcherny:

bcherny:
@ttyniwa PRs welcome!

PR #319, opened 2020-07-07
Allow to install package from GitHub

lundibundi:
This adds 'prepare' script that npm/yarn will run upon package
installation in node_modules. Previously installation from GitHub would
fail because dist is not committed to the repository but
'bin'->'json2ts' refers to files in it ('dist/src/cli.js') which
therefore fails the installation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "node": ">=10.0.0"
   },
   "scripts": {
+    "prepare": "npm run build",
     "build": "npm run clean && mkdir dist && npm run build:browser && npm run build:server",
     "build:browser": "browserify src/index.ts -s jstt -p tsify > dist/bundle.js",
     "build:server": "tsc -d",


bcherny:
Hey, thanks for the contribution! I'm not familiar with how GH packages work, but would this also mean we'd need to publish a GH package every time I publish a package to NPM?

Also, it would be great to avoid forcing people that install this package from NPM to have to run the `prepare` script, which might take a substantial amount of time to run.

lundibundi:
>  I'm not familiar with how GH packages work,

Oh, this is not for [GitHub packages](https://github.com/features/packages) but for installation from GitHub directly like `npm install github:bcherny/json-schema-to-typescript` (see https://docs.npmjs.com/cli/install or https://stackoverflow.com/questions/17509669/how-to-install-an-npm-package-from-github-directly).


> Also, it would be great to avoid forcing people that install this package from NPM to have to run the prepare script, which might take a substantial amount of time to run.


The `prepare` script is special and is run by both npm and yarn (ptal here for npm https://docs.npmjs.com/misc/scripts)
> prepare: Run both BEFORE the package is packed and published, on local npm install without any arguments, and when installing git dependencies (See below). This is run AFTER prepublish, but BEFORE prepublishOnly.

bcherny:
Got it, thanks for helping me understand @lundibundi. Will `prepare` run when folks install json-schema-to-typescript?

ie. If my project has the following package.json:

```json
{
  "dependencies": {
    "json-schema-to-typescript": "*"
  }
}
```

When I run `npm install`, will it run json-schema-to-typescript's `prepare` script after the install has completed?

lundibundi:
@bcherny yes it will run, though not after the install has completed but as a part of installing `json-schema-to-typescript` package right after all dependencies of `json-schema-to-typescript` were installed but before the package itself is set-up and "installed".

The installation is completely transparent for the user:
Screenshot from 2020-07-24 09-33-46

Current master:
Screenshot from 2020-07-24 09-33-59

aloisklink:
> Got it, thanks for helping me understand @lundibundi. Will `prepare` run when folks install json-schema-to-typescript?

Minor correction, it won't run when installing from NPM.

Some examples:

```bash
# the following will only install dependencies and NOT run the prepare script
npm install json-schema-to-typescript
npm install json-schema-to-typescript.tgz # generated via npm pack
```

```bash
# will install devDependencies and run `npm run prepare`, if it exists
npm install github:bcherny/json-schema-to-typescript
```

```bash
# will install devDependencies and run `npm run prepare`, if it exists
git clone https://github.com/bcherny/json-schema-to-typescript.git
cd json-schema-to-typescript
npm install

npm publish # will also run before npm publish
npm pack # will also run before npm pack
```

Most package managers, e.g. `npm`, `yarn` v1, `pnpm`, etc. all work the same way. `yarn` v2+ is the main exception to the rule, as it doesn't support `prepare`.

If you want to support `yarn` v2+ as well, you may want to look into `prepack`, which is almost the same as `prepare` except:
1. `prepack` is supported by `yarn` v2
2. Running `npm install` locally **won't** run `prepack`

bcherny:

PR #328, opened 2020-07-28
Properly generate unions that always have a given set of properties

coolreader18:
My specific use case for this is a Rust struct with a flattened enum field -> generated jsonschema with [schemars](https://docs.rs/schemars). e.g.

```rust
struct Thing {
    #[serde(flatten)]
    basic: BasicThing,
    #[serde(flatten)]
    specific: SpecificThing,
}
struct BasicThing {
    id: String,
    health: usize,
}
#[serde(tag = "type")]
enum SpecificThing {
    Foo { foo_type: FooType },
    Bar { bar_type: BarType, inventory: Inventory },
}
```
```typescript
export type Thing = (
    | { type: "Foo", foo_type: FooType }
    | { type: "Bar", bar_type: BarType, inventory: Inventory }
) & {
    id: string,
    health: number,
};
```


src/parser.ts
```diff
@@ -84,6 +84,46 @@ function parseNonLiteral(
 ) {
   log('blue', 'parser', schema, '<-' + typeOfSchema(schema), processed.has(schema) ? '(FROM CACHE)' : '')
 
+  function parseUnionType(params: JSONSchema[]) {
```
coolreader18, line 87: Not sure about putting this inside the function, but it seemed like the easiest way to get all of the parameters needed without making another function with basically the same signature as `parseNonLiteral`

bcherny (changes requested):
Thanks for the contribution, and sorry about the delay in taking a look. JSTT v10 handles this case more generally -- feel free to pull it and try it out.

I've also merged your test case from this PR to make sure v10 gives the same output as this PR: see b3b1d987dd0997c07e58f346dd3fd7a302799e14.

coolreader18:
Great, thank you for your work on this project!

Issue #329, opened 2020-07-31
Union Type has wrong type name if it in the 'root' level


TIS-OMiddle:
```
const schema = {
  type: ['number', 'string'],
  additionalProperties: false,
};

const code = await compile(schema, 'MyType')
```
the `code` shoule be:
```
export type MyType = MyType1 | MyType2;
export type MyType1 = number;
export type MyType2 = string;
```

actually:
```
export type MyType2 = MyType | MyType1;
export type MyType = number;
export type MyType1 = string;
```

I think the type "MyType" should be a union type. But in fact it is "MyType2". It's a feature or a bug?

PR #331, opened 2020-08-03
fix union types name

TIS-OMiddle:
Solve #329 .
```
const schema = {
  type: ['number', 'string'],
  additionalProperties: false,
};

const code = await compile(schema, 'MyType')
```

`standaloneName` method should be called before `parse` the `schema.type` ? 

The output should be:
```
export type MyType = MyType1 | MyType2;
export type MyType1 = number;
export type MyType2 = string;
```

Not very familiar with ava. The command `npm run tdd` just update `test/__snapshots__/test/test.ts.snap`, so the test is pass? 🤔


bcherny:
Fixed in e038017. Sorry I didn't see your PR earlier -- feel free to at-mention to bug me next time!

PR #336, opened 2020-09-08
fix(tests): Don't use snapshots for tests that provide an expected output

adrienjoly:
I noticed that `e2e/allOf.ts` exports an `output` value that is not used by the `e2e.ts` test runner, because it always relies on snapshots.

Explicit `output` values like that one are very useful – compared to snapshots – to follow the TDD development methodology: write a failing test with an explicit expectation, and only then implement the feature to make that test pass.

=> I propose to make `e2e` leverage that `output` value instead of snapshot, when it is provided by tests.


bcherny:
Thanks for the contribution! Instead, let's kill `output`, to keep things consistent (either all `output`, or all snapshots). Done in 7863d79.

PR #338, opened 2020-09-14
feat: multiple files to single output file

### LINKED ISSUE #272: [Feature] Add single file output for multi file input
#238 is adding the ability to specify multiple file input for this library. 

An additional feature would be to allow the user to specify a single file output for a multi file input and have everything be concatenated into the single file output.

Currently, a check is in place in the CLI to throw an error when the user attempts this operation.

tadhgmister:
Fixes #272

- adds new option `--exportMainOnly` (default to false) which causes all definition types to be not exported
- adds new option `usedNames` (not available to command line) which allows the same set of used names to persist over multiple calls to `compile` (this was main part needed to have single output file work correctly)
- when input is multiple files and output is stdout it now uses a continuous `usedNames` set so piping to a file works as expected
- when output is a single file and input is multiple files it writes it to the file similarly to outputting to stdout but it also overrides the banner for subsequent files to just be a separator.


bcherny:

PR #353, opened 2020-12-14
Add readOnly keyword support

bcaudan:
## Motivation

Generate typescript interface with `readonly` keyword by parsing corresponding JSON `readOnly` keyword value.
[JSON schema readOnly spec](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.9.4)

Input:
```JSON
{
  "title": "Example Schema",
  "properties": {
    "firstName": {
      "type": "string",
      "readOnly": true
    }
  }
}
```
Output:
```ts
  readonly firstName?: string
  [k: string]: unknown;
}
```
## Changes

- Parse readOnly value
- Generate readonly keyword
- Add specific test

## Questions

1. Should this behavior be behind a specific option?
2. About tests snapshot, I've updated it using ava `--update-snaphsots` option but it seems to have changed the order of the existing snapshots which makes the diff pretty unusable. how do you usually update tests snapshot?


src/parser.ts
```diff
@@ -353,6 +353,7 @@ function parseSchema(
     ast: parse(value, options, key, processed, usedNames),
     isPatternProperty: false,
     isRequired: includes(schema.required || [], key),
+    isReadOnly: !!value.readOnly,
```
bcherny, line 356: Let's try to be stricter here:

1. Add a normalizer step to add `readOnly: false` to each schema if it's not already defined
2. Update the type of `NormalizedJSONSchema` to reflect that `readOnly` is guaranteed to be defined
3. Update your changes to parser.ts to directly read `readOnly`, and avoid the double cast (`!!`)

test/e2e/readOnly.ts
```diff
@@ -0,0 +1,21 @@
+export const input = {
+  additionalProperties: false,
```
bcherny, line 2: What behavior would you expect if `readOnly` is used on the root schema? Seems like a good thing to support, though the spec is loosey goosey, as usual:

> An instance document that is marked as "readOnly for the entire document MAY be ignored if sent to the owning authority, or MAY result in an error, at the authority's discretion.

bcherny (changes requested):
This is a great start, thank you for the contribution! A couple of minor comments.

src/parser.ts
bcaudan: done in 6d4eec5

test/e2e/readOnly.ts
bcaudan: Indeed, I did not thought about this one.
From a typescript point of view, having `ReadOnly<Root>` could be interesting.
However, I am a bit puzzled by the spec, from my understanding, it is not the modifications of the values of the document that should be ignored, but the whole document. So, I am not sure to see a matching typescript concept here.
What do you think?

bcherny: We can land this PR as-is (once it's updated to use a normalizer step), and track `readonly` for root schemas in a separate issue.  How does that sound?

bcaudan: Sounds good to me

src/parser.ts
bcaudan: I've started to go this way but I'm struggling a bit to define [when to add this property to the schema or not](https://github.com/bcaudan/json-schema-to-typescript/commit/253a975883a4ebc2727298215d003d44bd0633ec#diff-f8b5c2873187ca697b26451706f58782775ac1c2656fc73e4fc1d8193febfe36R137).
Am I missing something?

BenoitZugmeyer:
Hello @bcherny.

@bcaudan and I are still waiting for your feedback so this PR can finally be merged. I see that in the meantime, [another PR with the same goal](https://github.com/bcherny/json-schema-to-typescript/pull/353) has been open. We would be interested in either of those PRs being released in `json-schema-to-typescript`.

Many thanks for your project!

src/parser.ts
bcherny: You can use the `isObjectSchema` and `isArraySchema` utils to help (see other normalizer steps for examples).

bcherny:
Closing in favor of https://github.com/bcherny/json-schema-to-typescript/pull/412/files. That said, this PR is more closely aligned with the approach I'd suggest -- feel free to revive this PR and request review if you'd like to continue work on it!

PR #361, opened 2021-01-08
[Feat] Memoize `getDefinition` parser function

G-Rath:
I've been working on schemas for Github webhook events with the goal of generating types from the schemas - we recently merged the initial schemas, and so I've gotten to the point of actually generating the types from the `schema.json` generated by running `npm run build:schema` on [this pr](https://github.com/octokit/webhooks/pull/252).

The `schema.json` is quite large - just shy of 30000 lines, and currently it takes ~120 seconds to build the `.d.ts` file.
While I know this is to be expected when running large schemas, I did some basic poking around and found that the bulk of the bottleneck is in the `getDefinitions` parser method.

That method had a jsdoc block with the comment "TODO: Memoize" so on a whim I decided to see how fast it'd be if I did just that.

Turns out it'll be 96.4% times faster, taking the time from ~120 seconds down to *~3 seconds* while still generating an identical schema.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1,5 +1,5 @@
 import {JSONSchema4Type, JSONSchema4TypeName} from 'json-schema'
-import {findKey, includes, isPlainObject, map, omit} from 'lodash'
+import {findKey, includes, isPlainObject, map, memoize, omit} from 'lodash'
 import {format} from 'util'
 import {Options} from './'
 import {typesOfSchema} from './typesOfSchema'
@@ -119,7 +119,7 @@ function parseNonLiteral(
   processed: Processed,
   usedNames: UsedNames
 ): AST {
-  const definitions = getDefinitions(getRootSchema(schema as any)) // TODO
+  const definitions = getDefinitionsMemoized(getRootSchema(schema as any)) // TODO
   const keyNameFromDefinition = findKey(definitions, _ => _ === schema)
 
   switch (type) {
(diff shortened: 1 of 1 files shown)


src/parser.ts
```diff
@@ -433,10 +434,9 @@ via the \`definition\` "${key}".`
 
 type Definitions = {[k: string]: LinkedJSONSchema}
 
-/**
- * TODO: Memoize
- */
-function getDefinitions(
+const getDefinitions = memoize(getDefinitionsInner)
+
+function getDefinitionsInner(
```
bcherny, line 439: Two small nits:

1. Would you mind using Lodash's `memoize` function, to avoid introducing the new dependency on `memoizee`?
2. A stylistic nit: could you avoid `getDefinitionsInner`, and instead inline the function? `Inner`, `Wrapper`, `Impl`, etc. often feel smelly to me. (Alternatively, call it `getDefinitions` and `getDefinitionsMemoized` to make the naming more meaningful.)

bcherny (changes requested):
Thanks for the contribution! A couple of minor nits.

src/parser.ts
G-Rath: Done! I've not inlined the function as it needs to call itself, but have improved the name and switched to using `lodash` :)

bcherny (approved):
Looks great!

bcherny:
Published v10.1.2

PR #362, opened 2021-01-11
fix: adjust type to remove need for `@types/lodash`

G-Rath:
Currently TS errors for this package if `@types/lodash` is not installed:

```
node_modules/json-schema-to-typescript/dist/src/types/JSONSchema.d.ts:1:23 - error TS2688: Cannot find type definition file for 'lodash'.

1 /// <reference types="lodash" />
                        ~~~~~~

node_modules/json-schema-to-typescript/dist/src/types/JSONSchema.d.ts:86:95 - error TS7016: Could not find a declaration file for module 'lodash'. '/c/Users/G-Rath/workspace/projects-oss/webhooks/node_modules/json-schema-to-typescript/node_modules/lodash/lodash.js'
 implicitly has an 'any' type.
  Try `npm i --save-dev @types/lodash` if it exists or add a new declaration (.d.ts) file containing `declare module 'lodash';`

86 export declare const getRootSchema: ((schema: LinkedJSONSchema) => LinkedJSONSchema) & import("lodash").MemoizedFunction;
```

I actually thought this was because of #361 but it's not - not sure why it didn't error in the past on the same project.

Originally I was going to add `@types/lodash` to `dependencies`, but realised I could explicitly type the export to make it not needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
   },
   "homepage": "https://github.com/bcherny/json-schema-to-typescript#readme",
   "dependencies": {
+    "@types/lodash": "^4.14.168",
     "@types/json-schema": "^7.0.6",
     "@types/prettier": "^2.1.5",
     "cli-color": "^2.0.0",
@@ -65,7 +66,6 @@
     "@types/cli-color": "^2.0.0",
     "@types/glob": "^7.1.3",
     "@types/is-glob": "^4.0.1",
-    "@types/lodash": "^4.14.165",
     "@types/minimist": "^1.2.1",
     "@types/mkdirp": "^1.0.1",
     "@types/mz": "^2.7.2",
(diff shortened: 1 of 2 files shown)


sempasha (approved):

bcherny (changes requested):
Hey, thanks for the contribution! Could we add the dependency on `@types/lodash` instead? Otherwise, without a way to add a test for this, this is likely to silently regress in the future (eg. next time someone refactors this function, and chooses to remove the unnecessary type annotation).

G-Rath:
@bcherny ugh that's a very good point 🤦‍♂️

A possible way to guard against that could be to export the `memoize` from a local file with a wrapper and comment explaining why, so that everything would use that `memoize`.

G-Rath:
(btw I've made some PRs to update the browser version to [the latest major](https://github.com/bcherny/json-schema-to-typescript-browser/pull/5) 🙂)

bcherny (approved):
Thanks for updating!

bcherny:
Published 10.1.3

utajum:
I am seeing this error with versions >= 15

Versions below 15 work fine

PR #364, opened 2021-01-29
chore: update `.npmignore` to exclude misc files

G-Rath:
The latest release (10.1.3) included a few misc files:

```
❯ npx npm@7 diff --diff json-schema-to-typescript@10.1.2 --diff json-schema-to-typescript@10.1.3 --diff-name-only
npx: installed 254 in 13.455s
dist/src/cli.js
package.json
yarn-error.log
package-lock.json
```

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,6 +2,7 @@
 .idea
 .vscode
 npm-debug.log
+yarn-error.log
 node_modules
 *.map
 example
@@ -11,3 +12,5 @@ circle.yml
 ARCHITECTURE.md
 CONTRIBUTING.md
 .editorconfig
+package-lock.json
+yarn.lock


.npmignore
```diff
@@ -11,3 +12,5 @@ circle.yml
 ARCHITECTURE.md
 CONTRIBUTING.md
 .editorconfig
+package-lock.json
+yarn.lock
```
bcherny, line 16: This should be checked in.

bcherny (changes requested):
LG! One minor request.

.npmignore
G-Rath: Why? This is used by `npm` when publishing, not for `git`.

Lock files like `yarn.lock` do nothing in published packages, so shouldn't be included.

bcherny: Ah, you're right. Good catch!

bcherny (approved):
LG!

PR #366, opened 2021-02-11
Include x-tsEnumNames and x-tsType as a option.

webmathias:
To agree with https://github.com/stoplightio/spectral/blob/0a8e0a6cbd53018d84352f749ba6ee2b71320498/src/rulesets/oas/schemas/schema.oas3.json where tsType and tsEnumNames are not allowed but x-tsEnumNames and x-tsType but no introduce break change.


bcherny (changes requested):
Hey, this seems reasonable to me! Instead of hard-coding `x-` properties everywhere, please add a pair of normalizer rules that:

1. Normalize `tsType` -> `x-tsType`
2. Normalize `tsEnumNames` -> `x-tsEnumNames`

Please add tests for these, then add an e2e test, and update the README to use `x-` properties.

Finally, please don't comment out tests. Can you help me understand why you commented it out?

webmathias:
@bcherny Sorry for the long time, I was a little busy, So I included the tests and the normalizations. I still have locally one error when I run 'npm run test' related with 'id?: (string | null)' but I believe to be. Let me know if you find anything that I can improve on the PR.

bcherny:
Hey @webmathias, WDYT about your approach vs. https://github.com/bcherny/json-schema-to-typescript/pull/405?

PR #368, opened 2021-02-22
Translate closed allOf -> closed intersection

jablko:
What do you think about translating e.g.
```TypeScript
{
  allOf: [
    {
      type: 'object',
      properties: {a: {type: 'string'}, b: {type: 'integer'}},
      additionalProperties: false,
      required: ['a', 'b']
    },
    {type: 'object', properties: {a: {enum: ['x']}}}
  ]
}
```
->
```TypeScript
    export type AllOf = {␊
      a: string;␊
      b: number;␊
    } & {␊
      a?: "x";␊
    };␊
```
vs.
```TypeScript
    export type AllOf = {␊
      a: string;␊
      b: number;␊
    } & {␊
      a?: "x";␊
      [k: string]: unknown;␊
    };␊
```
(removing `[k: string]: unknown;` in cases like this)

`{ a: "x", b: 1, additional: undefined }` is not an instance of the schema but is assignable to the latter type. It's not assignable to the former type (`Type '{ a: "x"; b: number; additional: undefined; }' is not assignable to type 'AllOf'. Object literal may only specify known properties, and 'additional' does not exist in type 'AllOf'.`).

This PR is a naive implementation. If any allOf children have `"additionalProperties": false` it removes `[k: string]: unknown;` from all of the allOf's (anonymous) children.

additionalProperties is handled [in `parseSchema()`](https://github.com/bcherny/json-schema-to-typescript/blob/eaac3684045976e0726a7fc1909d211b88dfc9f4/src/parser.ts#L403) which doesn't have access to the parent (the potential allOf context), so this implementation instead uses `"additionalProperties": false` to indirectly remove `[k: string]: unknown;` from children of a "closed" allOf.

TypeScript intersections don't quite line up with JSON Schema allOf (`{ "allOf": [<closed>, <open>] }` is closed whereas `<closed> & <open>` is open) but I think this change makes the type slightly more faithful to the schema.

This is useful if you want to extract the common parts of mostly similar, closed schemas. Currently if you do `{ "allOf": [{ "$ref": "#/components/schemas/common" }, { "type": "object", "properties": { "a": { "enum": ["x"] } } }] }`, where `#/components/schemas/common` has `"additionalProperties": false`, the type will nevertheless be open, but you can't add `"additionalProperties": false` to the second allOf item because that will preclude the common properties. The solution proposed here is to make the intersection more faithful to the schema in cases like this.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -124,11 +124,26 @@ function parseNonLiteral(
 
   switch (type) {
     case 'ALL_OF':
+      const closed = schema.allOf!.some(item => item.additionalProperties === false)
       return {
         comment: schema.description,
         keyName,
         standaloneName: standaloneName(schema, keyNameFromDefinition, usedNames),
-        params: schema.allOf!.map(_ => parse(_, options, undefined, processed, usedNames)),
+        params: schema.allOf!.map(item => {
+          const definitions = getDefinitionsMemoized(getRootSchema(item as any)) // TODO
+          const keyNameFromDefinition = findKey(definitions, _ => _ === item)
+          return parse(
+            closed &&
+              (item.additionalProperties === undefined || item.additionalProperties === true) &&
+              !standaloneName(item, keyNameFromDefinition, new Set<string>())
+              ? {...item, additionalProperties: false}
+              : item,
+            options,
+            undefined,
+            processed,
+            usedNames
+          )
+        }),
         type: 'INTERSECTION'
       }
     case 'ANY':


bcherny:
Hey, thanks for the contribution! Would you mind opening an issue first, to explain the motivation for this change a bit more? Also, tests are a must.

bcherny (changes requested):
^

bcherny:

PR #377, opened 2021-03-25
Replace dubious dependency `stdin` with `get-stdin`

albertodiazdorado:
* 'stdin' is a package consisting in one commit 8 years ago. There is no support,
there is no community, and it is not even easy to find the source code (you have
to download the tarball using the link from the yarn.lock, you will not find the
repository elsewhere)
* In contrast, `get-stdin` has active support and is more used by two orders of
magnitude
* I firmly believe that the world will be a better place when packages like `stdin`
slowly disappear from our dependency trees

PS: I love your package, keep it up!

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "@types/lodash": "^4.14.168",
     "@types/prettier": "^2.1.5",
     "cli-color": "^2.0.0",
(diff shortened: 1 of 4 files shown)


bcherny (approved):
Thanks for the contribution!

Issue #356, opened 2020-12-17
Broken generation with additionalProperties


isaacl:
```sh
> tsc --strictNullChecks Broken.d.ts
Broken.d.ts:11:3 - error TS2411: Property 'message' of type 'string | undefined' is not assignable to string index type 'string'.
```

Generated via v10.0.0:
```ts
...
export interface FormObject {
  message?: string;
  [k: string]: string;
}
```

Schema
```json
{
  "allOf": [{ "$ref": "#/definitions/FormObject" }],
  "$schema": "http://json-schema.org/draft-07/schema#",
  "definitions": {
    "FormObject": {
      "properties": {
        "message": {
          "type": "string",
          "maxLength": 4000
        }
      },
      "additionalProperties": {
        "type": "string",
        "maxLength": 100
      },
      "type": "object"
    }
  }
}
```

I left the maxLength properties in so you can see why I wrote the schema like this.

isaacl:
From [ts handbook](https://www.typescriptlang.org/docs/handbook/interfaces.html)
> While string index signatures are a powerful way to describe the “dictionary” pattern, they also enforce that all properties match their return type. This is because a string index declares that obj.property is also available as obj["property"].

isaacl:
So this is more broken than I thought.  Basically the index type in typescript enforces on all existing properties, while in json additionalProperties spec only applies to things not listed in properties.

isaacl:
additionalProperties probably needs to always translate to
```ts
[k: string]: unknown;
```

bcherny:
This is a bug, and we should find a better way to handle this case.

One option is to always emit index signatures as `[k: string]: T | undefined`. In your case, `[k: string]: string | undefined` would fix the TS error.

However, I'm not sure this is always desirable, and if it wasn't for the `message` property, this may impose an unreasonable burden on consumers, who then need to check every index property for `undefined` (note: if a consumer is using the `--noUncheckedIndexedAccess` TSC flag this may be what they want, but it's a burned for those consumers that don't use the flag).

A better approach may be to change the way we emit index signatures, to always emit them as an intersection type:

```ts
export type FormObject = {[k: string]: string} & {message?: string}

declare const x: FormObject

x.a.toUpperCase() // OK
x.message.toUpperCase() // Error: Object is possibly 'undefined'
```

I'd love input from others. Are there cases this might not handle? Examples welcome.

isaacl:
[...]

Issue #299, opened 2020-05-08
additionalProperties can create invalid typings


awlayton:
The following schema:

```json
{
  "properties": {
    "s": { "type": "string"}
  },
  "additionalProperties": {
    "properties": {
      "foo": {"type" : "string"}
    }
}
```

produces this typing:

```ts
/* tslint:disable */
/**
 * This file was automatically generated by json-schema-to-typescript.
 * DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,
 * and run json-schema-to-typescript to regenerate this file.
 */

export interface FooSchema {
  s?: string;
  [k: string]: {
    foo?: string;
    [k: string]: unknown;
  };
}
```

Which produces this `tsc` error:
```
foo.d.ts:9:3 - error TS2411: Property 's' of type 'string' is not assignable to string index type '{ [k: string]: unknown; foo?: string; }'.
```

bcherny:
Hey there! In this case, it's not clear to me what the expected behavior is here. It really is a contradiction that `s` must be both a string, and an object with a key` foo`.

Instead, consider expressing this using "oneOf", to make it clear that you want a shape that's one or the other, and not both. Something like:

```json
{
   "oneOf":[
      {
         "properties":{
            "s":{
               "type":"string"
            }
         },
         "additionalProperties":false
      },
      {
         "additionalProperties":{
            "properties":{
               "foo":{
                  "type":"string"
               }
            },
            "additionalProperties":false
         }
      }
   ]
}
```

Which generates:

```ts
export type Demo =
  | {
      s?: string;
    }
  | {
      [k: string]: {
        foo?: string;
      };
    };
```

awlayton:
The schema is not unclear though? `s` is always a string and any _other_ properties are objects with a `foo` key.

awlayton:
This is a fine type for what I want, which is exactly what my schema means:

```ts
/* tslint:disable */
/**
 * This file was automatically generated by json-schema-to-typescript.
 * DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,
 * and run json-schema-to-typescript to regenerate this file.
 */

export type FooSchema = {
  s?: string;
} & {
  [k: string]: {
    foo?: string;
    [k: string]: unknown;
  };
}
```

bcherny:
Maybe JSONSchema and TS have different semantics here:

```ts
export type S = {
  s?: string;
} & {
  [k: string]: {
[...]

PR #383, opened 2021-04-21
Export additionalProperties as intersection for complex types

fabrykowski:
A quick fix for #299, #356 and #402.

Given the schema

```json
{
  "title": "AdditionalProperties",
  "type": "object",
  "properties": {
    "foo": {
      "type": "string"
    }
  },
  "additionalProperties": {
    "type": "number"
  }
}
```

it is rendered by the library as
```typescript
export interface AdditionalProperties {
  foo?: string;
  [k: string]: number;
}
```

which results in the compiler error
```
TS2411: Property 'foo' of type 'string | undefined' is not assignable to string index type 'number'.
```

This PR renders the schema above as
```typescript
export type AdditionalProperties = {
  foo?: string;
} & {
  [k: string]: number;
};
```

which is valid TypeScript.


test/__snapshots__/test/test.ts.md
```diff
@@ -5682,6 +5691,142 @@ Generated by [AVA](https://avajs.dev).
     }␊
     `
 
+## realWorld.jsonschema.js
```
fabrykowski, line 5694: I have no idea, why this section is moved

jamesbohacek:
But it doesn't seem to actually fix the problem? Or am I doing something wrong?

The generated type is indeed valid TypeScript in the sense that it does compile. But if you try to create an object with it, it will still fail:

```ts
export type AdditionalProperties = {
  foo?: string;
} & {
  [k: string]: number;
};

const bar: AdditionalProperties = {
    foo: 'baz'
}

/*
Type '{ foo: string; }' is not assignable to type 'AdditionalProperties'.
  Type '{ foo: string; }' is not assignable to type '{ [k: string]: number; }'.
    Property 'foo' is incompatible with index signature.
      Type 'string' is not assignable to type 'number'.
*/
```

[TypeScript Playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcCCATDBLG2IB2AhgDYAKUEKs2wAznALxwDeAUHHAGYQQD8ALjh0YUbAQDmAbjYBfOADJWHOAG0A1kJFjJAXSEEArgFsARsCgzZMtgGNCIuKaJQhmHHkKkKVC3npMypycPBBCAOTOAF7hckA)

fabrykowski:
It is impossible to create such an object in TypeScript, but you can use it after a cast or assertion.

```typescript
export type AdditionalProperties = {
  foo?: string;
} & {
  [k: string]: number | undefined;
};

const bar = {
    foo: 'bar',
    baz: 42
} as unknown as AdditionalProperties;

const foo = bar.foo; // type: string | undefined
const baz = bar.baz; // type: number | undefined
const tmp = bar.tmp; // type: number | undefined
```

[TypeScript Playground](https://www.typescriptlang.org/play?ts=4.2.3#code/KYDwDg9gTgLgBDAnmYcCCATDBLG2IB2AhgDYAKUEKs2wAznALxwDeAUHHAGYQQD8ALjh0YUbAQDmAbjYBfOADJWHOAG0A1kJFjJAXSEEArgFsARsChwAPnEMEMwLuOAYZsmWwDGhEXFNFLZnZOTh4IIQByfygIgBoVTn8ALyEAFgAmOTgiBjt1AggAdwJshkwcPEJSCioLPHoPbwJfMKY-AIA6MKk4AHpehGRgLVFxCWtbe0dnDC8feGS26I7knv7BlAMTc0sbOwcnAhc55vgYYzAlzvOwNYGkTbgjMwsJ-emjjCA)

fabrykowski:
This is especially of interest, when a received JSON object – such as an HTTP request body – is parsed.

Michael-Hering (approved):

fabrykowski:
We need this in our code base – any chance of a merge?

Michael-Hering:
+1 - TS serverless files are a huge win for teams building serverless apps in TS

maximseshuk:
+1

luc-heyday:
+1

electblake:
+1

macsj200:
I would greatly appreciate this fix as I'm currently blocked by https://github.com/serverless/typescript/issues/27.

fabrykowski:
@bcherny – this PR would close multiple tickets. Would it be possible to initiate a review?

G-Rath:
@fabrykowski  (and potential reviewers) please keep in mind my comment here: https://github.com/bcherny/json-schema-to-typescript/issues/356#issuecomment-751144641

This change will be breaking as it prevents you from using declaration merging (which can only be done with `interface`s) which can be a very powerful tool with `additionalProperties`

fabrykowski:
@G-Rath – in the example you gave only the finished type `AdditionalProperties` is exported – which means that declaration merging would not be available. On the other hand exporting the `[...]KnownProperties`-interface for each type with `additionalProperties` seems a tad unwieldy and pollutes – in my opinion – the namespace.

I would argue that declaration merging is a very useful tool for DefinitelyTyped, allowing for example to define specific headers; should not the aim of this package be the opposite? If I have a JSON-schema and I validate a JSON-object (e.g. via [ajv](https://github.com/ajv-validator/ajv)) then I can safely cast the object to (or assert it to be) the generated TypeScript-type! (This is, at least, the way *I* use this package.)

If there is strong demand for it, I could add another optional compile option (`additionalPropertiesInterfaceSuffix`?) which, when not falsey, turnes on the export of `interface {TypeName}{additionalPropertiesInterfaceSuffix}`. I believe, however, that this should be a second PR, because this PR is "breaking" only in the most formal sense, since the currently generated interface does not compile anyways. (This PR still returns an `interface`, when no `additionalProperties` are present.)

G-Rath:
> in the example you gave only the finished type AdditionalProperties is exported ...
> exporting the [...]KnownProperties-interface for each type with additionalProperties seems a tad unwieldy and pollutes – in my opinion – the namespace.

Sure in a perfect world it'd be great to less stuff while still being able to have the flexibility but this is what the language requires us to do in order for this change to not be breaking + retain that flexibility.

> should not the aim of this package be the opposite
> I can safely cast the object to (or assert it to be) the generated TypeScript-type!

While neither are terrible alternatives, they are not without costs: casting can introduce bugs if not understood and maintained properly because the whole point of them is to have TypeScript ignore type violations like missing properties and type mismatches, and assertions can have a significant runtime cost depending on the size of the objects (which people may shortcut with approximation assertions, which is fine but then you're in the first tradeoff again).

> because this PR is "breaking" only in the most formal sense
> This PR still returns an interface, when no additionalProperties are present.

[Why is this being re-written as a `type` then?](https://github.com/bcherny/json-schema-to-typescript/pull/383/files#diff-34615f003eb49f90c3044b8fafebe3fe6b162e290b9b724082a6d72f58cfa4afL5942-L5944)

I'm definitely happier if it doesn't change interfaces with no `additionalProperties` are present, but I still think because there's a solution that allows you to have your cake and eat it too is better than this current one - splitting that out into _another_ config option IMO is bloating the library needlessly.

------

Please keep in mind that not everyone is validating their objects against the actual schemas at runtime - for example, over at https://github.com/octokit/webhooks we maintain schemas for all the webhook payloads which are used to generate the types that consumed by related packages; while the schemas are valid and can be used, they are not by the related octokit packages that are using the types. This is also the case for the rest of the GitHub API.

Declaration merging is very powerful here for the same reason its powerful over at DT: if we've made a mistake, or there's some new feature, and no one's gotten around to updating the schemas & types (we're pretty responsive but stuff happens), then people can easily amend the types using DM.

yourbuddyconner:
Bumping this PR because it appears that TS declaration file are straight up broken (still). I applied [this fix ](https://github.com/serverless/typescript/issues/27#issuecomment-779755994)to my local node_modules, but like... come on. 

How about less pedantic discussion about types and more fixy fixy?

barbunzel:
Would also love to have this merged!

test/__snapshots__/test/test.ts.md
```diff
@@ -1271,6 +1272,15 @@ Generated by [AVA](https://avajs.dev).
       [k: string]: unknown;␊
     };␊
     export type Header1 = ExampleXORExamples & SchemaXORContent;␊
+    export type Callback = {␊
+      /**␊
+       * This interface was referenced by \`Callback\`'s JSON-Schema definition␊
+       * via the \`patternProperty\` "^x-".␊
+       */␊
+      [k: string]: unknown;␊
```
bcherny, line 1280: Can we improve this output?

bcherny (changes requested):
Sorry for the delay.

This should be implemented as a normalizer rule, and not as a special case in the parser. The rule should be: If a schema specifies non-boolean `additionalProperties`, then rewrite the schema as an `allOf` of the schema (without `additionalProperties`) and the `additionalProperties`.

roman-shandurenko:
Can we get it merged, please? The issue has been open for 2 years and is still relevant and causes problems.

adambiggs:
Pretty please?

PR #384, opened 2021-04-21
Cross platform tests

Karel-Kroeze:
makes a number of small tweaks to the tests to allow them to successfully run on my platform, while hopefully not breaking them on other platforms;

- add double quotes to glob patterns
- sort list of paths by name to create a stable order (note: this is a breaking change, and fails with the current snapshot).
- use `path.join()` to generate paths

My apologies for the extra commits, I didn't initially notice I was continuing from a previous PR, so had to do some reverts.


bcherny:
Triggering CI run.

Karel-Kroeze:
note that failure was expected, given that the order of operations in some tests was changed to make sure they are stable across platforms. Tests should succeed after flushing the cached results.

gricey432:
Can this be rerun? I'm trying to contribute a PR from a windows machine and this is blocking me

Karel-Kroeze:
part of this PR was to sort the paths being tested alphabetically so that the order remains stable across platforms. This means tests comparing to a previous snapshot are expected to fail. It's been a while since I looked at this, but I recall being able to manually verify that all the failures were related to the path order. 

I'm not sure if that is still the case, so you should probably re-verify.

gricey432:
If the snapshot changes are expected, don't you need to include the changes in this PR? `yarn test --update-snapshots`

s-weigand:
In total the cross-platform testing has 3 problems where @Karel-Kroeze already fixed 2
1. The order of files is different depending on the platform (fixed here)
2. Using single quotes in the glob pattern doesn't resolve properly on windows (fixed here)
3. Using `path.join` and `path.normalize` is platform dependent (fixed in #464)

Luckily `path` also has `path.posix.join` and `path.posix.normalize`, which guarantee a posix path string on all platforms.
While windows and posix style paths work fine on windows, if you generate snapshots it will save windows style paths which then breaks tests on posix platforms. The solution is to only use posix style paths since they are [supported cross-platform](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.altdirectoryseparatorchar?view=net-6.0#remarks).

Damn I wish I had a dollar for each time windows path style was causing me problems 😅

bcherny:
https://github.com/bcherny/json-schema-to-typescript/pull/384

bcherny:
https://github.com/bcherny/json-schema-to-typescript/pull/384

PR #389, opened 2021-04-25
feat: Align node versioning and build to >10

robcresswell:
The package.json (which I'm taking here as the source of truth) lists Node >=10 as the minimum, but the `.nvmrc` lists `6.3.0` and the build output indicates very old Node.

This patch updates the `.nvmrc` and `tsconfig` to align around Node 10.

Note that Node 10 is EOL in 5 days as of writing this patch, so it might be worth going straight to 12, but thats up the maintainer(s)


robcresswell:
Closing; its been 2.5 months, so I'm guessing the author has other priorities 😇

PR #390, opened 2021-04-25
Add a LICENSE file

robcresswell:
This feels like an overly petty PR, but here's my reasoning; despite the `package.json` listing MIT as the license, there's no full declaration, and legal folks in this area can get a bit edgy about this. Rather than argue the merits of this either way, the easiest way forward seems to be to just add a small license.

From my point of view, this is mostly so I can keep using the library without any potential future issues, because I really like it

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Boris Cherny
+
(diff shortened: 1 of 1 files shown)


robcresswell:
Closing; its been 2.5 months, so I'm guessing the author has other priorities 😇

PR #394, opened 2021-05-16
prevent infinite recursion when validating schemas

cody-greene:
**The Problem:**
Summary: 
Infinite recursion is possible in src/utils.ts:mapDeep()
This will fix a few cases of "Uncaught RangeError: Maximum call stack size exceeded"

Detail:
Let's say I define a schema like this:
```javascript
const ListNode = {
  type: 'object',
  properties: {
    value: {type: 'string'},
    next: {$ref: '#'} // Circular reference
  }
}

// If I call compile() twice on the same schema object (with different options, for example)
import {compile} from 'json-schema-to-typescript'
const prettyContent = await compile(ListNode)
// The first call is fine

// But notice the ListNode object was modified (see $RefParser.dereference)
console.log(ListNode)
//{
//  type: 'object',
//  properties: { value: {type: 'string'}, next: [Circular *1] }
//}

const uglyContent = await compile(ListNode, {format: false})
// Uncaught RangeError: Maximum call stack size exceeded
// Uh-oh! There's infinite recursion happening.

// And of course I don't actually have to call compile() twice.
// I could just as well define my schema like this in the first place:
ListNode.properties.next = ListNode
```
**The Solution:**
Summary:
To fix this we can modify the recursive function in question to track seen objects and return early if one is seen second time (or more).

Detail:
**replace src/utils.ts:mapDeep() with crawl()**
- considering where mapDeep() is used, mutation is not necessary so we can remove the use of lodash's mapValues() and just iterate over Object.keys(obj)

**use a Set to detect reference cycles**
- it's fast and has broad browser support which includes IE11

**Final note:** Unfortunately json-schema-ref-parser has the exact same problem so we'll need to wait for https://github.com/APIDevTools/json-schema-ref-parser/pull/231 to merge before this is really solved.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -20,17 +20,17 @@ export function crawl(
   obj: any,
   fn: (value: object, key?: string) => void,
   key?: string, // internal
-  seen = new WeakMap<object, number>() // internal
+  seen = new Set<object>() // internal
 ): void {
   if (!seen.has(obj)) {
     if (isPlainObject(obj)) {
-      seen.set(obj, 1)
+      seen.add(obj)
       for (const key of Object.keys(obj)) {
         crawl(obj[key], fn, key, seen)
       }
       fn(obj, key)
     } else if (Array.isArray(obj)) {
-      seen.set(obj, 1)
+      seen.add(obj)
       for (const el of obj) {
         crawl(el, fn, key, seen)
       }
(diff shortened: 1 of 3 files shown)


bcherny:
Thanks for the contribution, and sorry for the delay in merging this in. Closing since https://github.com/bcherny/json-schema-to-typescript/commit/4837ef2940ea75d19808fd96dee3e389db2b9ad5 is now merged.

PR #397, opened 2021-06-01
Id or keyNameFromDefinition is used in name when title is not safe

boris-chernysh:
Hello!

In my schemas, all titles written in Russian, and now it means than all interfaces will be named as NoNameX.

I add the ability to use id or keyNameFromDefinition when title is unsafe.


src/parser.ts
```diff
@@ -300,9 +300,12 @@ function standaloneName(
   keyNameFromDefinition: string | undefined,
   usedNames: UsedNames
 ): string | undefined {
-  const name = schema.title || schema.id || keyNameFromDefinition
-  if (name) {
-    return generateName(name, usedNames)
+  const nameSources = [schema.title, schema.id, keyNameFromDefinition].filter(
+    source => typeof source !== 'undefined'
+  ) as string[]
```
bcherny, line 305: Can you simplify this code a bit by moving the `toSafeString` call into `standaloneName`?

```suggestion
  const name = [schema.title, schema.id, keyNameFromDefinition]
    .filter(Boolean)
    .map(toSafeString)
    .find(_ => _ != null)

  return generateName(name, usedNames)
```

bcherny (changes requested):
Sorry about the delay in getting back to you! This change seems reasonable. One small nit for code style. Please fix that and rebase, and I'll be happy to merge this!

bcherny:
Closing out stale PRs. If you'd like to revisit this, feel free to re-open.

PR #398, opened 2021-06-02
Allow falsey values in `const` properties

### LINKED ISSUE #393: Incorrect type for a `const` with an empty string value
Since #290 the `const` keyword is supported and acts as an alias for an `enum` with one value. This is great but it does not work correctly for a const that contains an empty string. Example:

Input:

```json
"foo": {
	"const": ""
}
```

Expected output:

```
foo: ""
```

Actual output:

```
foo: {
	[k: string]: unknown;
};
```

An enum containing a single value of an empty string does work as expected:

```json
"foo": {
	"enum": [
		""
	]
}
```

## What use is a const with an empty string?

I use these for fields where the value is required and always a string but can be empty, for example an optional URL. Of course it should be nullable/optional really but this is a legacy application.

```json
"url": {
	"oneOf": [
		{
			"type": "string",
			"format": "uri"
		},
		{
			"enum": [
				""
			]
		}
	]
},

johnbillion:
Fixes #393


bcherny:
Thanks for the bugfix!

bcherny (approved):
👍

PR #401, opened 2021-08-04
Add optional support for TypeScript's bigint

KtorZ:
- This change isn't changing any of the existing behavior by default.
  - When enabled however, it'll generate integers with no 'maximum' property as 'bigint'.

  Why?

  While JSON supports big integers in its specification, JavaScript does not by default. Thus, trying to deserialise a large integer into a JavaScript number is doomed to fail. Since relatively recently however, JavaScript (and therefore TypeScript), natively supports a 'bigint' type for numbers (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt), and there exists fixed parsers that can parse numbers into bigint directly.


src/typesOfSchema.ts
```diff
@@ -70,18 +71,30 @@ const matchers: Record<SchemaType, (schema: JSONSchema) => boolean> = {
...
     if (!isCompound(schema) && typeof schema.default === 'number') {
       return true
     }
     return false
   },
+  BIGINT(schema, options) {
+    if ('enum' in schema) {
+      return false
+    }
+    if (schema.type === 'integer' && options.enableBigInt && schema.maximum === undefined) {
```
bcherny, line 93: Should we also compare `maximum` with `Number.MAX_SAFE_INTEGER`?

test/e2e/bigint.2.ts
```diff
@@ -0,0 +1,21 @@
...
+  title: 'Example Schema -- BigInt Enabled',
+  type: 'object',
+  additionalProperties: false,
+  properties: {
+    secondsSinceDawnOfTime: {
+      type: 'integer'
+    },
+    reasonablySizedNumber: {
+      type: 'integer',
+      maximum: 42
```
bcherny, line 13: What happens when there's no `minimum`? Can you flesh out the behavior for negative bigints a bit more?

bcherny (changes requested):
I'm supportive of this change. Thanks for the high quality PR!

A couple of clarifying questions, and this will be ready to go after a rebase.

bcherny:
Closing out stale PRs. If you'd like to revisit this, feel free to re-open.

PR #411, opened 2021-09-21
Add support for nullable primitives

goodoldneon:
Picking up where #312 left off.

This PR adds a normalizer that converts `nullable: true` primitive types into `anyOf` types.


src/normalizer.ts
```diff
@@ -145,6 +145,30 @@ rules.set('Transform const to singleton enum', schema => {
...
+
+    const copiedSchema = {...schema}
+
+    // This stuff should not be in `anyOf`.
+    delete copiedSchema.id
+    delete copiedSchema.nullable
+
+    // This stuff will be in `anyOf` instead.
+    delete schema.format
+    delete schema.type
```
bcherny, line 160: Can you remove all of these `delete`s? Unclear why you need them (156 is already included in 150; 155+159+160 I'm not sure what they're here for).

src/normalizer.ts
```diff
@@ -145,6 +145,30 @@ rules.set('Transform const to singleton enum', schema => {
...
+
+    schema.anyOf = [
+      copiedSchema,
+      // @ts-expect-error
```
bcherny, line 164: Can you rm this?

bcherny (changes requested):
Thanks for the contribution! A few minor suggestions.

src/normalizer.ts
goodoldneon: I get a TS error without it:

```
Type '{ type: "null"; }' is missing the following properties from type 'LinkedJSONSchema': additionalProperties, [Parent]ts(2739)
```

bcherny: Cool! Can you find a way to fix the error? eg. to add an empty `additionalProperties`, and to add a `Parent` field?

klarstrup:
Bump 😭

drewcorlin1:
I'd love to be able to use this feature as well. Is there anything I can do to help this PR get merged?

austin-garner:
Bump

src/normalizer.ts
goodoldneon: @bcherny 

I think we need all of these deletes. After normalizing, we no longer want `nullable`, `format`, or `type` to appear in the root of the schema. That's what the `anyOf` is replacing.

For example, the following schema:

```json
{
  "id": "foo",
  "type": "integer",
  "format": "int32",
  "nullable": true
}
```

Should be normalized to this:

```
 {
  "id": "foo",
  "anyOf": [
    {
      "type": "integer",
      "format": "int32"
    },
    {
      "type": "null"
    }
  ]
}
```

Notice how `nullable` is gone, and `type` and `format` have moved into the `anyOf`

goodoldneon:
Sorry, y'all. I don't write TypeScript at my new job (unfortunately 😭) so I forgot about this PR. I just responded to a comment the maintainer had, so hopefully I'll be able to push this over the finish line soon

lucas-inflab:
I really need to make this change, is there anything I can do such as new pull request?

bcherny:
Closing in favor of https://github.com/bcherny/json-schema-to-typescript/pull/522

PR #423, opened 2021-11-17
Fix strictIndexSignature with standalone types

mrcljx:
Removes the safe-naming reapplication against `generateType` return values
which might cause invalid typescript definition files when `strictIndexSignature` is used.

```diff
interface MyInterface {
- [k: string]: LeafUndefined; // error TS2304: Cannot find name 'LeafUndefined'.
+ [k: string]: Leaf | undefined;
}
```

Please let me know if you think a different codepath needs to be fixed.
But to me it looks like `generateType` calls `generateRawType` which already has:

```typescript
  if (hasStandaloneName(ast)) {
    return toSafeString(ast.standaloneName)
  }
```


bcherny:
Published 13.0.0.

PR #427, opened 2021-12-12
write JSDoc tags for unkown properties

### LINKED ISSUE #237: "Not expressable in TypeScript" as jsdoc comments
For the "Not expressible in TypeScript:" list in the README, I am wondering whether you would be interested in providing default output whereby such items could be expressed as jsdoc comments in the generated output (and if not, at least ensuring that the proposed custom templating in #9 would have enough information provided to it, such as the current schema object, that users could do so themselves).

Even the nestable items, `not`, `oneOf`, and `dependencies`, should, I think, be representable by generating any necessary (non-exported) type structures and referencing them within the comments, e.g., `@not AutoGeneratedType1` and then building the necessary (non-exported) interface `AutoGeneratedType1`. While TypeScript won't use this information outside of allowing it to pass on through, the information will at least be preserved.

I think having a default behavior would be ideal for both facilitating use and reuse of validation keywords within TypeScript comments and facilitating round-tripping with the likes of typescript-to-json-schema converters.

If you are open to the idea, there is the question of which jsdoc tags to use. One could use one single custom jsdoc tag (e.g., `@ts maximum 100`, `@ts minimum 50`) or different custom jsdoc tags (e.g., `@ts-maximum 100`, `@tx-minimum 50` or just `@maximum 100`, `@minimum 50`). I believe the existing generic standard jsdocs tag like `@description`, `@classdesc`, or `@summary` would not be ideal candidates here for this kind of info. (FWIW, while some parsers like [comment-parser](https://github.com/yavorskiy/comment-parser/blob/master/parser.js#L104) assume a tag title to be comprised of `\w`, looking at [jsdoc source](https://github.com/jsdoc/jsdoc/blob/aa0b6c1bfa3ab1aea10e579f9d6168fcb8ad5332/lib/jsdoc/doclet.js#L124), I see any non-whitespace should be allowed in a tag title, so the hyphen example could apparently be supported and not requiring an underscore.

elazarcoh:
Allow user to generate custom/unknown jsonschema properties as JSDoc tags. 
It should (help? fully?) fixes #237
Example:
```jsonc
// person.json
{
  "title": "Person",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string"
    },
    "lastName": {
      "type": "string"
    },
    "address": {
      "type": "array",
      "items": {
        "type": "string"
      },
      "default": [
        "123 Main St",
        "Anytown",
        "CA",
        "90210"
      ]
    },
    "age": {
      "description": "Age in years",
      "type": "integer",
      "minimum": 0
    },
    "hairColor": {
      "enum": ["black", "brown", "blue"],
      "type": "string"
    }
  },
  "required": ["firstName", "lastName"]
}
```
```ts
async function generate() {
  writeFileSync('person.d.ts', await compileFromFile('person.json', {jsdocTags: ["description", "minimum" , "default"]}))
}

generate()
```
generates:
```ts
/* tslint:disable */
/**
 */

export interface Person {
  firstName: string;
  lastName: string;
  /**
   * @default ["123 Main St","Anytown","CA","90210"]
   */
  address?: string[];
  /**
   * @description "Age in years"
   * @minimum 0
   */
  age?: number;
  hairColor?: "black" | "brown" | "blue";
  [k: string]: unknown;
}
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -61,6 +61,10 @@ export interface Options {
    */
   unknownAny: boolean
   /**
+   * JSDoc tags to parse from the schema. Will be prepended to each generated type in the comment block (as `@tag [tagValue]`).
+   */ 
+  jsdocTags: string[]
+  /**
    * [$RefParser](https://github.com/BigstickCarpet/json-schema-ref-parser) Options, used when resolving `$ref`s
    */
   $refOptions: $RefOptions
@@ -90,7 +94,8 @@ export const DEFAULT_OPTIONS: Options = {
     useTabs: false
   },
   unreachableDefinitions: false,
-  unknownAny: true
+  unknownAny: true,
+  jsdocTags: [],
 }
 
 export function compileFromFile(filename: string, options: Partial<Options> = DEFAULT_OPTIONS): Promise<string> {
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -121,7 +121,7 @@ function parseNonLiteral(
 ): AST {
   const definitions = getDefinitionsMemoized(getRootSchema(schema as any)) // TODO
   const keyNameFromDefinition = findKey(definitions, _ => _ === schema)
-  const jsdoc: Record<string, any> = options.jsdocTags.map(prop => [prop, schema[prop]]).reduce((acc, [key, val]) => (val !== undefined ? { ...acc, [key]: val } : acc), {});
+  const jsdoc = parseTags(schema, options.jsdocTags);
   const jsdocOrComment = isEmpty(jsdoc) ? { comment: schema.description } : { jsdoc: jsdoc };
 
   switch (type) {
@@ -308,6 +308,10 @@ function standaloneName(
   }
 }
 
+function parseTags(schema: { [key: string]: any }, tags: string[]): { [key: string]: any } {
+  return tags.map(prop => [prop, schema[prop]]).reduce((acc, [key, val]) => (val !== undefined ? { ...acc, [key]: val } : acc), {});
+}
+
 function newInterface(
   schema: SchemaSchema,
   options: Options,
@@ -317,7 +321,7 @@ function newInterface(
   keyNameFromDefinition?: string
 ): TInterface {
   const name = standaloneName(schema, keyNameFromDefinition, usedNames)!
-  const jsdoc: Record<string, any> = options.jsdocTags.map(prop => [prop, schema[prop]]).reduce((acc, [key, val]) => (val !== undefined ? { ...acc, [key]: val } : acc), {});
+  const jsdoc = parseTags(schema, options.jsdocTags);
   const jsdocOrComment = isEmpty(jsdoc) ? { comment: schema.description } : { jsdoc: jsdoc };
   return {
     ...jsdocOrComment,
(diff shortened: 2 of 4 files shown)


src/index.ts
```diff
@@ -60,6 +60,10 @@ export interface Options {
    * Generate unknown type instead of any
    */
   unknownAny: boolean
+  /**
+   * JSDoc tags to parse from the schema. Will be prepended to each generated type in the comment block (as `@tag [tagValue]`).
+   */ 
+  jsdocTags: string[]
```
bcherny, line 66: Instead of making this a configurable option, what do you think of emitting an `@tag [tagValue]` line for each property that's a valid JSONSchema property, but not supported by JSON Schema to TypeScript? We know what these are, so there's no need to make consumers manually pass it in.

*utils.ts* would be a good place to put a `Set` of those properties, which you can then check against in `parseTags`.

src/parser.ts
```diff
@@ -121,11 +121,13 @@ function parseNonLiteral(
 ): AST {
   const definitions = getDefinitionsMemoized(getRootSchema(schema as any)) // TODO
   const keyNameFromDefinition = findKey(definitions, _ => _ === schema)
+  const jsdoc = parseTags(schema, options.jsdocTags);
+  const jsdocOrComment = isEmpty(jsdoc) ? { comment: schema.description } : { jsdoc: jsdoc };
```
bcherny, line 125: What happens when a schema has both unsupported properties, and a description? (hint: we should emit both)

src/types/AST.ts
```diff
@@ -23,15 +24,21 @@ export type AST =
 
 export interface AbstractAST {
   comment?: string
+  jsdoc?: Record<string, any>
```
bcherny, line 27: Nit: Please keep capitalization consistent, and avoid `any` like fire.

```suggestion
  jsDoc?: {[k: string]: unknown}
```

src/types/AST.ts
```diff
@@ -23,15 +24,21 @@ export type AST =
...
   type: AST_TYPE
 }
 
 export type ASTWithComment = AST & {comment: string}
+export type ASTWithJSdoc = AST & { jsdoc: Record<string, any> }
 export type ASTWithName = AST & {keyName: string}
 export type ASTWithStandaloneName = AST & {standaloneName: string}
 
+export function hasJsdoc(ast: AST): ast is ASTWithJSdoc {
+  return 'jsdoc' in ast && ast.jsdoc != null && !isEmpty(ast.jsdoc)
```
bcherny, line 39: No need for `isEmpty`, just mimic the checks in `hasComment`.

bcherny (changes requested):
Thanks for the contribution, and sorry for the delay -- this is a strong start!

Please fix the suggestions, and add tests for at least:

1. Emitting a schema that has 1 unsupported property
2. Emitting a schema that has a few unsupported properties
3. Emitting a schema that has both a description and an unsupported property
4. Emitting a schema that has both a description and a few unsupported properties
5. Emitting a schema that has a property that references a definition that has an unsupported property
6. (Any other important cases you can think of)

src/index.ts
elazarcoh: Hmm... In fact I'm using this library to generate a custom JSON (VSCode contribution points json), not specifically a JSON Schema.
So, I agree that having default values for known JSON Schema properties is a good idea, but I still want to have a way to generate from any custom tag. 
Does that make sense?

bcherny:

Issue #335, opened 2020-08-31
Feature request: Configuration option to disable generation of `[k: string]: unknown`


binaryunary:
Hi,

Would you consider adding a config option to disable generation of `[k: string]: unknown;` if `additionalProperties` is not explicitly set to false?

I understand that technically the current behavior is correct as it follows the spec but `[k: string]: unknown;` allows referencing any arbitrary properties and makes it difficult to catch errors at compile time.

tadhgmister:
you definitely want to complain to who ever owns the schema telling them to add `additionalProperties: false` since users of the schema that don't use typescript will face the same issue.

bcherny:
Agreed with @tadhgmister -- while this may seem appealing, it's an easy way to introduce subtle bugs where codegenned typings behave differently for different languages. We want to stick as closely as possible to the JSON-Schema spec to make the compiler predictable and reliable.

Leaving this open for now to collect more feedback.

johnbillion:
Agreed that this is a problem that should be fixed in the JSON schema, not at the point where its transformed into TS interfaces. If the JSON schema you're referencing is used for validation then the "any arbitrary properties" you refer to are perfectly valid.

Karel-Kroeze:
while I agree that forcing `additionalProperties` to be false should never be the default, having the option to do so through an option for overly permissive schemas would be extremely handy.

dhoepelman:
I would like to vote for this as a config option.
I'd even argue that it should be the default as typescript programs can always have excess properties at runtime¹.

`additionalProperties: false` has some caveats for use - adding a field in a newer schema becomes a breaking change - and you don't always control the schema.

You should still exclusively use known properties in your programs. `[k: string]: unknown` makes that more error-prone.

¹ Typescript uses duck typing. [It can prevent additional properties in special cases](https://www.typescriptlang.org/docs/handbook/interfaces.html#excess-property-checks).

Example ([Playground Link](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAbgQwDYFcCmMC8MDeAoGQmBALhgHIFyAaAogIzPPvLwF888AzFMYKAJbgYXABQCwABxRQyOYmWgAnCQHMYbAJS46hJWigolYEcghoA3O05jEqNNoD0jmGBAw0SpSCVA))
```typescript
const value = {
    a: 'a',
    b: 'b'
}

function f(input: { a: string }) {
    return false;
}

f(value) // no error
```

[...]

PR #445, opened 2022-05-11
Add additionalPropertiesSignature option

nbrahms:
👋 Howdy 👋 

Implements #335.

There is some debate on #335 as to the propriety of this change. My intent
with this PR is to eliminate effort as a roadblock to implementation of this
change pending resolution of said debate.

Of course, I have a horse in the race, but defer to the maintainers as to whether
they want to incorporate this behavior.


test/__snapshots__/test/test.ts.md
```diff
@@ -5682,6 +5706,142 @@ Generated by [AVA](https://avajs.dev).
     }␊
     `
 
+## realWorld.jsonschema.js
```
nbrahms, line 5709: Looks like the`tdd` snapshot ordering is non-deterministic

bcherny (changes requested):
Thanks for the PR! Superseded by https://github.com/bcherny/json-schema-to-typescript/commit/59747b1f086216a2e42efcf207884e98dd8fdc77

PR #447, opened 2022-05-15
Optimize optimizer to avoid deep JSON serialization (fix #422)

### LINKED ISSUE #422: RangeError: Invalid String Length at JSON.stringify()

bcherny:
This diff introduces a "peeking" behavior in the optimizer, to look ahead at the types we would have generated for a given AST node, and use that to de-dupe types before emission.

This means our optimizer can now more efficiently de-dupe AST nodes (avoiding general-purpose JSON serialization), significantly improving performance in cases where schemas emit large unions of complex types.

This supercedes the bandaid we were using before, where we tried to "safely" JSON stringify to compare AST nodes. It also supercedes https://github.com/bcherny/json-schema-to-typescript/pull/424, as well as the more complex hashcode-based approach I prototyped in https://github.com/bcherny/json-schema-to-typescript/pull/446.


jmikrut:
Hey @bcherny — was this ever released to NPM? 

It looks like a new version hasn't been published yet. Let me know if I'm missing something.

Thank you!

PR #460, opened 2022-06-09
inline object POC

goloveychuk:
Hey. This is POC of generating inline nested types.
If it's something that could be merged, I'll finish PR: put it under flag `generateInlineTypes: true` and add tests. 
Yeah, I know it's more limiting, but lib consumer sometimes needs to have inline type. 
Also I'll throw an error if smth could be not generated (e.g. recursive types).

### DIFF
--- a/src/formatter.ts
+++ b/src/formatter.ts
@@ -5,5 +5,9 @@ export function format(code: string, options: Options): string {
   if (!options.format) {
     return code
   }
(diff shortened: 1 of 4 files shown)


bcherny:
Hey, I have no clue what you're trying to do here. Can you please open an issue and document the motivation for your proposal before re-opening this PR? Thanks!

elmage:
@bcherny I find myself also in need of this, sadly.

Issue #466, opened 2022-07-07
Schema definitions are generated multiple times when setting a property description


queengooborg:
This change was found somewhere between 10.1.5 and 11.0.1.  When generating types, I am noticing that typedefs are generated twice in certain scenarios, particularly when referencing a schema definition.

Schema:
```js
// This schema has been simplified to include only the relevant portions of the schema
{
  "$schema": "http://json-schema.org/schema#",

  "definitions": {
    /* ... */
    "status_block": {
      "type": "object",
      "properties": {
        "experimental": {
          "type": "boolean",
          "description": "A boolean value that indicates whether this functionality is intended to be an addition to the Web platform. Set to false, it means the functionality is mature, and no significant incompatible changes are expected in the future."
        },
        "standard_track": {
          "type": "boolean",
          "description": "A boolean value indicating whether the feature is part of an active specification or specification process."
        },
        "deprecated": {
          "type": "boolean",
          "description": "A boolean value that indicates whether the feature is no longer recommended. It might be removed in the future or might only be kept for compatibility purposes. Avoid using this functionality."
        }
      },
      "required": ["experimental", "standard_track", "deprecated"],
      "additionalProperties": false
    },

    "compat_statement": {
      "type": "object",
      "properties": {
        /* ... */
        "status": {
          "$ref": "#/definitions/status_block",
          "description": "An object containing information about the stability of the feature.",
          "tsType": "StatusBlock"
        }
      },
      /* ... */
    },
}
```

Before:
```ts
export interface CompatStatement {
  /* ... */
  /**
   * An object containing information about the stability of the feature.
   */
  status?: StatusBlock;
}

/**
 * This interface was referenced by `CompatDataFile`'s JSON-Schema
 * via the `definition` "status_block".
 */
export interface StatusBlock {
  /**
   * A boolean value that indicates whether this functionality is intended to be an addition to the Web platform. Set to false, it means the functionality is mature, and no significant incompatible changes are expected in the future.
   */
  experimental: boolean;
  /**
   * A boolean value indicating whether the feature is part of an active specification or specification process.
   */
  standard_track: boolean;
  /**
   * A boolean value that indicates whether the feature is no longer recommended. It might be removed in the future or might only be kept for compatibility purposes. Avoid using this functionality.
   */
  deprecated: boolean;
}
```

After:
```ts
/**
 * An object containing information about the stability of the feature.
 */
export type StatusBlock = StatusBlock;
export interface CompatStatement {
[...]

PR #473, opened 2022-07-29
fix: allow descriptions on enum refs

### LINKED ISSUE #472: Adding a description to a $ref object referencing an enum inlines the enum
Perhaps related to #470, but not the same issue. Here, I'm seeing that when a `$ref` pointing to a defined enum has a `description`, the enum gets inlined.

```json
{
  "$definitions": {
    "shared": {
      "enum": ["a", "b"]
    }
  },
  "properties": {
    "first": {
      "$ref": "#/$definitions/shared",
      "description": "A first property."
    }
  },
  "title": "Example Schema",
  "type": "object"
}
```

Note that if the `description` is removed from `properties.first`, the `export type Shared = "a" | "b";` gets printed as expected.

## Actual

```ts
  /**
   * A first property.
   */
  first?: "a" | "b";
}
```

## Expected

```ts
export type Shared = "a" | "b";

  /**
   * A first property.
   */
  first?: Shared;
}
```

JoshuaKGoldberg:
Fixes #472.


src/parser.ts
```diff
@@ -474,3 +474,10 @@ const getDefinitionsMemoized = memoize(getDefinitions)
 function hasDefinitions(schema: LinkedJSONSchema): schema is JSONSchemaWithDefinitions {
   return 'definitions' in schema
 }
+
+function definitionMatchesSchema(definition: LinkedJSONSchema, schema: LinkedJSONSchema): unknown {
+  if ('enum' in definition && 'enum' in schema) {
+    return definition.enum === schema.enum
+  }
+  return definition === schema
+}
```
JoshuaKGoldberg, line 483: I'm not sold that this is a particularly good solution 🙃 it feels weird to hardcode the case for enums. But I'm not familiar enough with the codebase to have any deeper insights.

dhvector: I noticed the same on #466, but not related to enums. In their case, they had a property like this:

```
"status": {
  "$ref": "#/definitions/status_block",
  "description": "..."
}
```

See my [comment](https://github.com/bcherny/json-schema-to-typescript/issues/466#issuecomment-1200571984) as to why.

For a more universal fix that goes beyond just enums, I wonder about:
* copy the original parsed schema content
* cull certain fields from any subschemas that use `$ref`
* _(Now the $ref resolver can make the $ref referentially equal to what it's referencing, so that `json-schema-to-typescript` doesn't generate a new type (or inline, depending on what version you're using?))_
* During code generation, reference the original json for things like `description`, which shouldn't affect type relationships, but should certainly affect generated doc strings.

wdyt?

JoshuaKGoldberg: Honestly, although that sounds reasonable, I'm very new to this project and haven't done a deep dive into it. So if you say that's a good fix then I believe you! 😄

PR #483, opened 2022-08-22
add tsName

safareli:
This is more feature request  than a working pr.

Would be very nice to have `tsName` prop allowed and used as shown in this PR. if this is something you like please let me know and will add tests


bcherny:
Have you see the existing https://github.com/bcherny/json-schema-to-typescript/pull/456? This seems like a dup.

PR #487, opened 2022-09-21
perf: remove blank character at the beginning of string result

sangle7:
When I try to read a file encoding in utf-8 BOM,  it return a string with a "blank" as first character,  I looked into the code and found this error:

```
SyntaxError: Unexpected token ﻿ in JSON at position 0
    at JSON.parse (<anonymous>)
    at compileFromFile  (....)
```
Thus, I believe we should add a trim to remove the blank characters

Related issue: https://github.com/nodejs/node/issues/20649

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -115,7 +115,7 @@ export function compileFromFile(filename: string, options: Partial<Options> = DE
     }
   )
   const schema = Try<JSONSchema4>(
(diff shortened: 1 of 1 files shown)


bcherny (changes requested):
Thanks for the contribution! Can you please add a test? Then I'd be happy to merge this.

PR #495, opened 2022-11-16
Add MIT License

YairHalberstadt:
This looks like an amazing project, but a license is required in order for us to use it. Would you mind adding an explicit MIT license?

Thanks!


bcherny (approved):
Thanks for the contribution!

bcherny:
Published v11.0.4

PR #497, opened 2022-11-30
Make index properties optional

wbt:
In https://github.com/trufflesuite/truffle/issues/5746 there is an example of a schema where several properties are specified in addition to a patternProperty.  In converting this to TypeScript types, this library marks the properties as optional (i.e. allowing undefined) while the index property is not marked as allowing `undefined`. This produces TS compilation errors because the type of a specified property (which allows for `undefined`) is not assignable to the type of the index property (which does not allow for `undefined`).  
This seems like it would likely affect more than just that one example, so fixing it here seems like the best place. 
This PR gets part of the way there but doesn't fully fix the issue, leaving a TODO where someone who knows this codebase better than I do can pick up much more effectively.  
I would welcome other contributions and/or maintainer edits that get this farther along.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -364,10 +364,10 @@ function parseSchema(
     // additionalProperties is not set, and there is only a single
     // value definition, we can validate against that.
     singlePatternProperty = !schema.additionalProperties && Object.keys(schema.patternProperties).length === 1
-
     asts = asts.concat(
       map(schema.patternProperties, (value, key: string) => {
         const ast = parse(value, options, key, processed, usedNames)
+        //TODO: Modify ast here to add union with undefined, or add undefined to union
         const comment = `This interface was referenced by \`${parentSchemaName}\`'s JSON-Schema definition
 via the \`patternProperty\` "${key}".`
         ast.comment = ast.comment ? `${ast.comment}\n\n${comment}` : comment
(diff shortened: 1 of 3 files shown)


bcherny (changes requested):
Thanks for the attempted contribution. Feel free to create an issue and explain the problem you're running into + what you'd expect. Let's discuss in an issue before jumping into the implementation. Thanks.

wbt:
There is an open issue linked to above, and a description of the problem.  Here again is the core: 

> this library marks the properties as optional (i.e. allowing undefined) while the index property is not marked as allowing `undefined`. This produces TS compilation errors because the type of a specified property (which allows for `undefined`) is not assignable to the type of the index property (which does not allow for `undefined`).

Separately,
> Let's discuss in an issue before jumping into the implementation. Thanks.

Your policy seems to be the opposite of the one in TS itself where starting with the PR is in practice generally required so as to be able to demonstrate feasibility of a fix: otherwise, if it seems hard, the issue will be summarily closed (and future re-reports properly closed as duplicates) making it highly unlikely a reported issue without a PR will get addressed in the future. 

My workaround is a script that modifies the output file from this process to add `| undefined` to the index type.  It's not a robust generic solution and doesn't solve the problem for anyone other than myself and my direct colleagues.  This PR is opened in the hopes that its contributions might lead to something helpful for others.

chrissyw247:
+1 to this request! Also ran into this issue when I renamed the from from `.d.ts` to `.ts`.

bcherny:

wbt:
I still think something like this PR would be a good idea for the reasons identified, but would require some collaboration instead of requiring 100% of the knowledge of this code-base to come from an outsider.

PR #503, opened 2023-01-12
Add support for boolean JSON schemas

Natoandro:
Add support for boolean JSON schemas
(https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-boolean-json-schemas),
through the option `booleanSchemas`.


notaphplover:
Looks great, I just have some minor suggestions:

1. In general I consiider this should be a fix. I mean, boolean JSON Schemas are about to be correctly parsed thanks to this fix. For this specific reason, I consider no `booleanSchemas` option should be included.

2. ~~Maybe `NEVER` is a bit to typescriptish for the parsing context, whose domain is JSON Schema. Maybe there's a less coupled name for our case.~~ Nah, I cannot find a better name tbh.

README.md
```diff
@@ -96,6 +96,7 @@ See [server demo](example) and [browser demo](https://github.com/bcherny/json-sc
 | unreachableDefinitions | boolean | `false` | Generates code for `definitions` that aren't referenced by the schema. |
 | strictIndexSignatures | boolean | `false` | Append all index signatures with `\| undefined` so that they are strictly typed. |
 | $refOptions | object | `{}` | [$RefParser](https://github.com/BigstickCarpet/json-schema-ref-parser) Options, used when resolving `$ref`s |
+| booleanSchemas | boolean | false | Generate respectively `any` and `never` types for boolean schema values `true` and `false` |
```
bcherny, line 99: This behavior should be the default -- it doesn't need to be configurable. Let's remove this option.

src/typesOfSchema.ts
```diff
@@ -35,6 +35,9 @@ const matchers: Record<SchemaType, (schema: JSONSchema) => boolean> = {
   ALL_OF(schema) {
     return 'allOf' in schema
   },
+  NEVER(schema) {
+    return (schema as any) === false
```
bcherny, line 39: Can you avoid this type assertion, and instead ensure upstream that the type for `schema` declares that it can be a `JSONSchema | boolean`?

(If you're not comfortable doing that, I can do it as a follow-up after we merge this PR.)

src/parser.ts
```diff
@@ -37,6 +37,20 @@ export function parse(
   usedNames = new Set<string>()
 ): AST {
   if (isPrimitive(schema)) {
+    if (options.booleanSchemas) {
+      if (schema === true) {
+        return {
+          keyName,
+          type: 'ANY'
```
bcherny, line 44: Careful to respect the `unknownAny` option here. Could you make sure your tests cover what happens when the option is set to `true`/`false`?
```suggestion
          type: options.unknownAny ? 'UNKNOWN' : 'ANY'
```

src/parser.ts
```diff
@@ -37,6 +37,20 @@ export function parse(
   usedNames = new Set<string>()
 ): AST {
   if (isPrimitive(schema)) {
+    if (options.booleanSchemas) {
+      if (schema === true) {
```
bcherny, line 41: Please move this new logic into the `parseLiteral` function.

test/e2e/options.booleanSchemas.ts
```diff
@@ -0,0 +1,10 @@
+export const input = {
+  properties: {
+    always: true,
+    never: false
```
bcherny, line 4: Can you expand these tests to:

1. Verify the output when the `unknownAny` option is `true` and `false` (as I noted above).
2. Verify the output when a sub-schema is `true` or `false` (try to cover at least a nested property and a definition).
3. Verify the output when the top-level schema itself is `true` or `false` (ie. `export const input = true`, `export const input = false`).

bcherny (changes requested):
Looking good -- let's take one more pass at this, then get it merged!

src/parser.ts
notaphplover: In addition, this is sort of buggy. Enabling boolean requires the distinction between parsing boolean literals and boolean schemas. Enum schemas with boolean values requires parsing the boolean value as a literal, as an example.

bcherny:

PR #506, opened 2023-02-01
Add option to restore previous behavior of dereferencing schema input

gamedevsam:
Our team has leveraged this behavior to work with the dereferenced schema, we've been avoiding updating this library for a while as a result of this change.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -75,7 +75,11 @@ export interface Options {
   /**
    * Generate unknown type instead of any
    */
-  unknownAny: boolean
+  unknownAny: boolean,
+  /**
+   * Mutate the input schema by dereferencing JSON schema refs (this restores old behavior of altering the input object)
+   */
+  defererenceInputSchema: boolean
 }
 
 export const DEFAULT_OPTIONS: Options = {
@@ -104,7 +108,8 @@ export const DEFAULT_OPTIONS: Options = {
     useTabs: false
   },
   unreachableDefinitions: false,
-  unknownAny: true
+  unknownAny: true,
+  defererenceInputSchema: false
 }
 
 export function compileFromFile(filename: string, options: Partial<Options> = DEFAULT_OPTIONS): Promise<string> {
@@ -139,7 +144,7 @@ export async function compile(schema: JSONSchema4, name: string, options: Partia
   }
 
   // Initial clone to avoid mutating the input
-  const _schema = cloneDeep(schema)
+  const _schema = _options.defererenceInputSchema ? schema : cloneDeep(schema)
 
   const {dereferencedPaths, dereferencedSchema} = await dereference(_schema, _options)
   if (process.env.VERBOSE) {
--- a/test/testIdempotence.ts
+++ b/test/testIdempotence.ts
@@ -20,6 +20,13 @@ export function run() {
     t.deepEqual(before, SCHEMA)
   })
 
+  test('compile() should mutate its input if defererenceInputSchema option is true', async t => {
+    const after = cloneDeep(SCHEMA)
+    const before = cloneDeep(after)
+    await compile(after, 'A', { defererenceInputSchema: true })
+    t.notDeepEqual(before, after)
+  })
+
   test('compile() should be idempotent', async t => {
     const a = await compile(SCHEMA, 'A')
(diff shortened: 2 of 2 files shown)


bcherny (changes requested):
I'd like to support your use case, but need more context about what you're trying to achieve. Can you please link an issue to explain why you need this mutation?

gamedevsam:
There is no issue I can give you, the project I'm working on is closed source.

At the core of the issue, is that I need to work with the de-referenced schema after generating types from it (I generate JS from the JSON schema).

Without this option, I will need to first reference a large schema to generate types, and then do it again to generate my JS. The previous behavior resulted in a more efficient flow.

Our schemas are quite large, so de-referencing it takes several seconds. This is why we've refused to upgrade since version `10.1.5`.

bcherny:

gamedevsam:
I gave you my reasoning why this PR is valuable, are you not convinced of its value? I can't share source code as I mentioned before.

My team is stuck on v10.1.5, I would like to upgrade it at some point, but the lack of this option is holding us back.

gamedevsam:
@bcherny I'd like to revisit this issue and explain the reasoning for why I believe this option is valuable.

Part of generating TypeScript from the JSON Schema, involves dereferencing and linking the schema. This converts a schema that has `$ref` elements into a fully normalized schema without any refs.

I leverage this flattened schema to generate JavaScript that allows me to work with the schema in my code, so in addition to generating types from the schema (using your library), I leverage the fact that your library normalizes the schema, to convert the normalized schema in JavaScript / TypeScript that can be consumed by clients, for example:

```ts
export const entity_debug_data = {
  get: () => ({
    $id: 'entity/debug_data',
    $schema: 'http://json-schema.org/draft-07/schema#',
    title: 'DebugData',
    type: 'object',
    required: ['type', 'components', 'status'],
    properties: {
      type: { enum: ['debug_data'] },
      components: {
        type: 'object',
        required: ['type', 'data'],
        properties: { type: { type: 'string' }, data: { type: 'object', required: [], additionalProperties: true } },
        additionalProperties: false
      },
      id: { type: 'string' },
      owner_id: { type: 'string' },
      status: { type: 'string', enum: ['active', 'inactive'], default: 'active' },
      created_at: { type: 'string', format: 'iso-datetime' },
      updated_at: { type: 'string', format: 'iso-datetime' }
    },
    additionalProperties: false
  }),
  new: (override?: unknown) => {
    const original = { type: 'debug_data', components: { type: '', data: {} }, status: 'active' } as const;
    return (override != null ? merge(original, override) : original || {}) as typeof original;
  },
  validate: async (payload: unknown, options?: import('jsonschema').Options) => {
    const { validate } = await import('jsonschema');
    return validate(payload, entity_debug_data.get(), options);
  }
};
```

This code is fully generated from JSON schema, which is the source of truth for my data models.

Does this make sense?

While I could use `@apidevtools/json-schema-ref-parser` to dereference the schema again after generating types from it, it's redundant since your library literally already does this, and all I need to enable this functionality is an optional config to avoid cloning the schema before dereferencing it (which I recognize is a good practice, and I'm fine with it, just give me the option to use the very useful footgun imo).

bcherny:
@gamedevsam Thanks for sharing your use case. We won’t be supporting mutating schemas — this indeed was a bug, as you suggested.

> While I could use @apidevtools/json-schema-ref-parser to dereference the schema again after generating types from it

This is the right approach. If you want to avoid waste, feel free to dereference your schema first before passing it in.

If you need to do it as part of JSTT, feel free to fork the library, or offer a more compelling use case.

gamedevsam:
I respect your decision (slippery slope supporting one off use cases), but I'll just probably stick with v10.1.5 forever (and fork it if I run into some weird edge case). It's not a bug, it's a feature for me :-P

gamedevsam:
I caved in and de-referenced the schema before passing into the updated version of this library. Appreciate your input and support for this library and json-schema-ref-parser <3

gamedevsam:
To be honest part of the reason I hadn't done it before is because another library I was using (json-schema-instanciator) didn't handle `const` values correctly, so updating your lib broke theirs. But I opened a PR to fix that on that side: https://github.com/tomarad/JSON-Schema-Instantiator/pull/36

Issue #508, opened 2023-02-06
Parse Enum values as literals


notaphplover:
### Description

Enum values should be parsed as literals as the spec states.

#### Expected behavior:

The following JSONSchema:

```JSON
{
  "title": "Enum with JSON literal",
  "enum": [ { "type": "string" } ]
}
```

should be transpiled into the following typescript model:

```typescript
export type EnumWithJSONLiteral = {type: "string"};
```

### Current behavior:

The following JSONSchema:

```JSON
{
  "title": "Enum with JSON literal",
  "enum": [ { "type": "string" } ]
}
```

is transpiled into the following typescript model:

```typescript
export type EnumWithJSONLiteral = string;
```

### Context

Consider the specs as references:

- https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00#section-5.20
- https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-01#section-6.23
- https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-00#section-6.1.2
- https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-6.1.2
- https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-02#section-6.1.2
- https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-01#enum

notaphplover:
@bcherny I just created #509 to fix this issue :)

bcherny:
Published json-schema-to-typescript-12.0.0

PR #509, opened 2023-02-06
Update parser to parse enum values as literals

notaphplover:
### Changed
- Updated parser to parse enum values as literals. Consider #508 as reference


test/e2e/enum.4.ts
```diff
@@ -0,0 +1,4 @@
+export const input = {
+  title: 'Enum with JSON literal',
+  enum: [{type: 'string'}]
```
bcherny, line 3: Can you add some more members to this enum? eg.

- Another object
- An array
- A string
- A nested object/array

bcherny (changes requested):
Thanks for fixing this long-standing bug! Let's beef up the tests a bit, then ship this as part of the next major version.

test/e2e/enum.4.ts
notaphplover: Sure thing @bcherny, I'll do it tomorrow :)

PR #515, opened 2023-02-23
Fix boolean json schemas

notaphplover:
### Added
- Added `NEVER` schema type.

### Changed
- Updated parser to correctly parse boolean JSON schemas.


notaphplover:
Hey @bcherny, this should do the trick to support boolean JSON schemas. Do you have any test cases in mind?

I added a couple of cases but I'm open to add / remove more :)

src/parser.ts
```diff
@@ -37,6 +38,10 @@ export function parse(
   usedNames = new Set<string>()
 ): AST {
   if (isPrimitive(schema)) {
+    if (isBoolean(schema)) {
+      return parseBooleanSchema(schema, keyName, options)
+    }
+
     return parseLiteral(schema, keyName)
```
notaphplover, line 45: tbh I don't know what to do here. In theory we should omit anything at this point instead of parsing the literal, but I don't want to introduce too many changes in this PR

src/types/JSONSchema.ts
```diff
@@ -126,6 +127,10 @@ export const getRootSchema = memoize((schema: LinkedJSONSchema): LinkedJSONSchem
   return getRootSchema(parent)
 })
 
+export function isBoolean(schema: LinkedJSONSchema | JSONSchemaType): schema is boolean {
```
notaphplover, line 130: Maybe this is sort of an overengineered approach, but I couldn't think in anything better

src/parser.ts
```diff
@@ -103,6 +108,20 @@ function parseAsTypeWithCache(
...
+    return {
+      keyName,
+      type: options.unknownAny ? 'UNKNOWN' : 'ANY'
+    }
+  } else {
+    return {
+      keyName,
+      type: 'NEVER'
+    }
+  }
```
bcherny, line 122: Nit

```suggestion
  }
  return {
    keyName,
    type: 'NEVER'
  }
```

src/parser.ts
bcherny: Could be a little cleaner to move the `parseBooleanSchema` into `parseLiteral`, since `true`/`false` are literals.

bcherny (changes requested):
This looks great -- thanks for the contribution! Couple of minor suggestions. Could you also please add test cases for top-level `true`/`false` schemas?

ie.

```
export const input = true
export const input = false
```

src/parser.ts
notaphplover: I think this would be a mistake. Parsing `true` and `false` as schemas is not the same as parsing them as values. Depending on the context, `true` and `false` are schemas or values.

notaphplover:
Hey @bcherny, when adding the root boolean schema case, it seems the ref parser can not dereference it, it throws an error instead. I can have a look at it, but seems the parser is expecting an object instead of a boolean value.

bcherny:
Published 13.0.0.

PR #516, opened 2023-02-24
fix: generate same title one types

### LINKED ISSUE #510: Identical properties with same `title` generate multiple types
Example: A rectangle schema with two corners. Both corners are the same type and most importantly share a `title`.

I know this isn't an ideal schema (should ref in the points), but sometimes you just get bad data from somewhere.
```js
import { compile } from 'json-schema-to-typescript';
import * as fs from 'node:fs';

const schema = {
    type: "object",
    title: "rectangle",
    properties: {
        p1: {
            type: "object",
            title: "point",
            properties: {
                x: "number",
                y: "number"
            },
            required: ["x", "y"]
        },
        p2: {
            type: "object",
            title: "point",
            properties: {
                x: "number",
                y: "number"
            },
            required: ["x", "y"]
        }
    },
    required: ["p1", "p2"]
}

fs.writeFileSync("dupe-example.ts", await compile(schema, undefined, { additionalProperties: false, bannerComment: "" }))
```

this turns into
```ts
export interface Rectangle {
  p1: Point;
  p2: Point1;
}
export interface Point {
  x: "number";
  y: "number";
}
export interface Point1 {
  x: "number";
  y: "number";
}
```
which is correct, I guess? after all, I'm putting garbage in and get garbage out. I'd prefer to get something like
```ts
export interface Rectangle {
  p1: Point;
  p2: Point;
}
export interface Point {
  x: "number";
  y: "number";
}
```
but I'm not sure if that's in the scope of this project. I saw #456, but as far as I understand it, that only allows changing the interface name _before_[^tsname] the incrementing number is appended, not after (so the same as just setting the title).

Ideally, I'd want a way to hard override the generated interface name. "Trust me, this is a `Point`, don't complain about duplicates"[^dedupe].
As a more complex alternative, this package could "detect" identical properties (with or without matching titles) and reuse the interfacces generated from those. Again, this might be out of scope - we're starting to shift from parse+generate into ✨ magic ✨ 

Any advice? I'd also be fine with a hacky workaround, can't be worse than regexing through my .ts file[^ignore] :^)

[^dedupe]: I'm already committing a kind of crime here: I take in lots of schemas and compile them individually, all into one file. I run [ts-dedupe](https://www.npmjs.com/package/ts-dedupe) on that file which gets rid of interfaces with the same name. This helps a lot when multiple schemas define a `"title": "point"` somewhere, but can't duplicates within a schema by itself. ts-dedupe only compares interface names, not their contents/definitions.

[^tsname]: see https://github.com/bcherny/json-schema-to-typescript/pull/456/files#diff-abe2818e9c29aadf84401cb7ef8286c7cbce9253eba09f6fa52e22ca4e57b17eR303-R305 - [`generateName`](https://github.com/bcherny/json-schema-to-typescript/blob/137f6a7cad767ff86d80018717b45bbc7762212c/src/utils.ts#L184) is called after the override and adds a counter

[^ignore]: As a final note: I don't have to have neat types, since I can just create my own `Point` interface and use that everywhere. After all, typescript will let identical types from different definitions "fit onto" each other. As long as nobody looks at the schemas' type definitions, the 10 different `PointXX` interfaces aren't noticable

ianfuin:
If two schemas are (a) identical and (b) have the same explicit title, then emit just one type to represent the schema

fix #510

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ See [server demo](example) and [browser demo](https://github.com/bcherny/json-sc
 | unreachableDefinitions | boolean | `false` | Generates code for `$defs` that aren't referenced by the schema. |
 | strictIndexSignatures | boolean | `false` | Append all index signatures with `\| undefined` so that they are strictly typed. |
 | $refOptions | object | `{}` | [$RefParser](https://github.com/BigstickCarpet/json-schema-ref-parser) Options, used when resolving `$ref`s |
+| sameExplicitTitle | boolean | `false` | Generate same title one types. If two schemas are (a) identical and (b) have the same explicit title, then emit just one type to represent the schema |
 ## CLI
 
 A CLI utility is provided with this package.
--- a/src/generator.ts
+++ b/src/generator.ts
@@ -29,9 +29,17 @@ export function generate(ast: AST, options = DEFAULT_OPTIONS): string {
   ) // trailing newline
 }
 
-function declareEnums(ast: AST, options: Options, processed = new Set<AST>()): string {
+function declareEnums(ast: AST, options: Options, processed = new Set<AST>(), usedNames = new Set<string>()): string {
   if (processed.has(ast)) {
     return ''
+  }
+
+  if (options.sameExplicitTitle && typeof ast.standaloneName !== 'undefined') {
+    if (usedNames.has(ast.standaloneName)) {
+      return ''
+    }
+
+    usedNames.add(ast.standaloneName)
   }
 
   processed.add(ast)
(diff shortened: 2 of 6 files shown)


jemand771:
hi @ianfuin and thanks for giving this a shot :D

this mostly works for me, but I think I've found a small regression:
```json
{
    "anyOf": [
        {
            "title": "host",
            "enum": [
                "example.com",
                "localhost"
            ]
        },
        {
            "title": "ip",
            "enum": [
                "123.123.123.123"
            ]
        }
    ],
    "type": "string",
    "title": "server"
}
```
would previously turn into
```ts
export type Server = Server1 & Server2;
export type Server1 = Host | Ip;
export type Host = "example.com" | "localhost";
export type Ip = "123.123.123.123";
export type Server2 = string;
```
but is now
```ts
export type Server = Server;
```
all while calling `compile` with the following options:
```js
compile(
        schema,
        undefined,
        {
            bannerComment: "",
            additionalProperties: false,
            strictIndexSignatures: true,
            sameExplicitTitle: sameExplicitTitle,
        }
    );
```
The example above is inspired by an `anyOf` schema with `{"format": "ipv4"}` and `{"format": "hostname"}`, which this lib doesn't seem to support - the `enum` keys are just there to make it at least a tiny bit realistic, but leaving them out gives a similar result.

I can "solve" this by removing either `type` or `title` from the outmost object. I can understand removing the title as it's redundant, but the type should just work like that[^factoring]

[^factoring]: https://json-schema.org/understanding-json-schema/reference/combining.html#id11

jemand771:
hm, I can also trigger the `type Foo = Foo;` case using just
```js
import { compile } from 'json-schema-to-typescript/dist/index.js';

console.log(await compile(
    {
        title: "host",
        tsType: "Host"
    },
    undefined,
    {
        bannerComment: ""
    }
));
```
which yields
```ts
export type Host = Host;
```
BUT that also happens without `sameExplicitTitle` set. maybe I'm just using `tsType` wrong, but it still feels related.
possibly also related to #466 (`export type StatusBlock = StatusBlock;` there)

bcherny (changes requested):
I might be misreading your code, but where do you check that two schemas are identical? With your code, it seems like if two schemas share the same name they will share the same emitted type, even if their properties are completely different.

This also needs more tests, and it should be the default behavior (not behind a flag).

bcherny:

PR #521, opened 2023-04-12
fix: allow other properties on schemas with refs whilst still generating a shared type

bradzacher:
Fixes https://github.com/bcherny/json-schema-to-typescript/issues/472.

This isn't QUITE right yet because the comment isn't on the property...


bcherny:
Closing out stale PRs.

PR #534, opened 2023-06-29
Respect declareExternallyReferenced for non-interface types

### LINKED ISSUE #525: `declareExternallyReferenced` should be respected for non-interface types, too
Consider the following JSON schema files:

`foo.schema.json`

```json
{
  "title": "Foo",
  "oneOf": [
    {
      "$ref": "./bar.schema.json"
    }
  ]
}
```

`bar.schema.json`

```json
{
  "title": "Bar",
  "type": "string"
}
```

Compiling `foo.schema.json` with the option `declareExternallyReferenced` set to false would lead to `bar.schema.json` being present in the generated code:

```js
import {compileFromFile} from 'json-schema-to-typescript'
import {ensureDir} from 'fs-extra'

ensureDir('dist')
  .then(() => {
    const schema = 'src/foo.schema.json';

    return compileFromFile(schema, {
      declareExternallyReferenced: false
    });
  })
  .then((code) => {
    console.log(code);
  });
```

```ts
/* eslint-disable */
/**
 */

export type Foo = Bar;
export type Bar = string;
```

Except if I'm missing the purpose of  `declareExternallyReferenced `, the generated code should be:

```ts
/* eslint-disable */
/**
 */

export type Foo = Bar;
```

alexmojaki:
Fixes #525


bcherny (changes requested):
This won't quite work as expected, I think -- what happens in this case?:

```ts
export const input = {
  type: 'object',
  definitions: {
    a: {
      title: 'a',
      type: 'string'
    }
  },
  properties: {
    a: {
      $ref: '#/definitions/a',
    }
  },
}
```

Can you spec out how this should behave a bit better? As designed, I'd expect `declareExternallyReferenced: false` to still emit a declaration for `A` in this case, since `A` was declared locally.

We might want a way to pipe through whether a dereference was external or not. See resolver.ts.

bcherny:
Closing out stale PRs.

PR #535, opened 2023-07-02
Support nullable

mdmower:
- Add transform for `nullable: true` to include `null` as a possible type for any of the following schema types: `const`, `enum`, `type`.
- Log a warning when `const` is not set to `null` but schema is marked `nullable: true`
- Log a warning when `enum` does not contain `null` but schema is marked `nullable: true`
- Add tests and update snapshot

This is a continuation of #522 by @jackfrankland to finish adding support for `nullable: true` in schemas.


src/normalizer.ts
```diff
@@ -222,6 +223,29 @@ rules.set('Transform const to singleton enum', schema => {
...
+  }
+
+  delete schema.nullable
+
+  // schema.const has already been converted to a single-value enum by rule "Transform const to singleton enum", so it
+  // does not need to be handled here.
+  if (schema.enum) {
+    if (!schema.enum.includes(null)) {
+      schema.enum.push(null)
+      log('yellow', 'normalizer', 'enum should include "null" when schema is nullable', schema)
```
mdmower, line 238: 1. Need to remove quotes around `null` in the message since the value in the enum should not have quotes
2. Unsure whether this is the level of log desired, or if it would be better to use `error()` from utils.ts

src/normalizer.ts
```diff
@@ -222,6 +223,29 @@ rules.set('Transform const to singleton enum', schema => {
...
 
+rules.set('Transform nullable to null type', schema => {
+  if (schema.nullable !== true) {
+    return
+  }
+
+  delete schema.nullable
+
+  // schema.const has already been converted to a single-value enum by rule "Transform const to singleton enum", so it
+  // does not need to be handled here.
```
mdmower, line 234: Should `schema.const` handling be added in this normalizer in case rules are reordered in the future and "Transform const to singleton enum" no longer precedes this one?

src/normalizer.ts
```diff
@@ -222,6 +223,29 @@ rules.set('Transform const to singleton enum', schema => {
   }
 })
 
+rules.set('Transform nullable to null type', schema => {
+  if (schema.nullable !== true) {
+    return
+  }
+
+  delete schema.nullable
```
mdmower, line 231: Do we need to delete the `nullable` property if it has any other value than `true`? e.g. if it is `false`, I don't imagine we'd actually want to put effort into searching for `null` in enum/const or `"null"` in type/anyOf/oneOf, but we might still want to delete the property? Or is that unnecessary?

mdmower:
@bcherny - Would you be willing to review?

src/utils.ts
```diff
@@ -209,6 +209,13 @@ export function error(...messages: any[]): void {
   console.error(getStyledTextForLogging('red')?.('error'), ...messages)
 }
 
+export function warning(...messages: any[]): void {
+  if (!process.env.VERBOSE) {
+    return console.warn(messages)
+  }
+  console.warn(getStyledTextForLogging('yellow')?.('warning'), ...messages)
```
zrosenbauer, line 216: This will always log. If you look at the `log` function below it actually returns `void` when its NOT verbose.

Since this is just a warning I recommend the following change:
```suggestion
  if (!process.env.VERBOSE) {
    return
  }
  console.warn(getStyledTextForLogging('yellow')?.('warning'), ...messages)
```

zrosenbauer (commented):
Other than the `warn` looks good to me. 

Once you make the update or comment back & it makes sense to resolve happy to approve.

src/utils.ts
mdmower: This was written to mimic https://github.com/bcherny/json-schema-to-typescript/blob/6adcad98ac28334f1e8cd932e412c64dde4205f5/src/utils.ts#L205-L210 

I don't pretend to understand that choice and it's not my aim to question it. I'm just following precedence here.

mdmower:
@bcherny - I still hold hope that this PR could move forward. Is there anything I can do to improve its chances of merge? Maybe you have an opinion on zrosenbauer's review above? The logging was added based on [this comment](https://github.com/bcherny/json-schema-to-typescript/pull/522#discussion_r1184789927) and jackfrankland's subsequent reply.

mdmower:
Maintainer is not interested. Abandoning.

PR #541, opened 2023-07-20
Add support for Snowplow's self-describing schemas

bjornhenriksson:
Many companies use [Snowplow Analytics](https://snowplow.io/) to manage their events. Snowplow uses a specific JSON schema structure (iglu) that involves a concept of [self-describing schemas](https://snowplow.io/blog/introducing-self-describing-jsons/). These schemas use a `self` property to add information about what version or name etc the event is. In extension this means that snowplow schemas typically don't include `$id` or `title` fields, as you can see in this official example: https://github.com/snowplow/iglu-example-schema-registry/blob/master/schemas/com.example_company/example_event/jsonschema/1-0-0. 

In the above example project's file structure (that many companies follow) we can't fall back on the file name either since it will be represented as numbers (not valid interface names). 

This PR adds support to transform the self.name field to the schema title via the normalizer, which should be favourable when using snowplow/iglu schemas. 

## For reviewers
- In this "naive" implementation I'm only checking if the `self` property exists and assume that's a value that contains "name" and then delete the whole "self" property. I'm happy to revise this if you think it will collide with other peoples use of this library, if its too unsafe or if you'd like to decorate further information from the self etc. Open to suggestions :)) 
- I've added a snapshot and a normalizer test
- Updated `yarn.lock` when running `npx yarn` (with the node version specified in the `.nvmrc`)


bcherny:
Thanks for the contribution, but I'm not sure we want to support this non-standard behavior. Here's what I would suggest:

1. Fix the upstream Snowplow schema generation to emit a proper `$id` field. Without this, JSTT and many other tools will not work correctly.
2. Feel free to open an issue (and submit a PR) for emitting valid interface names when interfaces are numbers. I think #539 might cover it, actually.
3. Finally, feel free to open an issue in this repo to track support for `self`. If there's a lot of interest, we can resurrect this PR.

Issue #551, opened 2023-09-06
Support ESM usage of module


julrich:
We're currently working on some ESM-based tooling (we're going pure ESM there), and in the process of migrating existing code into that cleaned-up code base we've stumbled upon some challenges with `json-schema-to-typescript`.

Currently the project only targets CJS. If there's interest in supporting both ESM and CJS, I'd be open to do a PR for that. We'd need to work on that anyways (less preferably as a fork).

One main thing that would be interesting to me in that context is the use of the `json-schema-ref-parser` fork. There seems to be some instability as a consequence of maintainer changes over in that code base. Initially ESM-support was added there (https://github.com/APIDevTools/json-schema-ref-parser/commit/96b805ada1427e05df7e97a0b5751f9ee0decbe5), but the latest rewrite seems to have completely removed that again (https://github.com/APIDevTools/json-schema-ref-parser/commit/a5b3946fbb62683ab69e3747a8893014591726af#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519).

For the use in `json-schema-to-typescript` the fork (https://github.com/bcherny/json-schema-ref-parser) seems to be exactly in between those last big rewrite changes. Is this just coincidence? My impression was that the main fix on the fork was about the node version being wrong in some in between-releases of `json-schema-ref-parser`.

If the intention would be to go back to a non-forked version of `json-schema-ref-parser` as soon as possible, this might be a blocker. Because those last rewrite changes over there removed everything about ESM apparently. If continuing to work of the already existing fork, ESM there should be a pretty easy fix. It's just line 3 over here that needs to be changed to "export default $RefParser" in my first initial, local testing.

julrich:
Added some initial work for this here: https://github.com/bcherny/json-schema-to-typescript/pull/552

bcherny:
Can you please better explain the issue you are running into when consuming JSTT in your ESM project? My understanding is most bundlers have an interop layer, so if something isn't working, it's more likely a misconfiguration on your end than something we should fix in JSTT.

If I'm missing something, a good start would be a quick demo of the issue you're running into. Could you throw a demo in a new repo and post a link?

julrich:
I'll give it a try: We're not actually using a bundler, only TypeScript for compilation. Not sure how you'd arrive at a miconfigured bundler.

When adding `json-schema-to-typescript` to the project, importing (e.g.) `compile` and building the application nets the following error:

```
[...]

PR #552, opened 2023-09-10
WIP: ESM + CJS dual support

julrich:
Some initial work to enable ESM usage of the package. Also keeps CJS for backwards compatibility.
The changes are in part based on this article by Anthony Fu: https://antfu.me/posts/publish-esm-and-cjs

Additionally, tsup is used to dual-bundle: https://github.com/egoist/tsup

Test suite still runs, and is all green.

Still though, more testing is needed for:
- [ ] CJS usage... ESM works, which is what we're using, but CJS is kind of untested (though I think the test suite should access the CJS version, and all the tests still pass)
- [ ] Browser usage. I think this would be best tested through https://github.com/bcherny/json-schema-to-typescript-browser? Haven't dove in there yet, though


bcherny:
Thanks for the contribution, but I'm a little confused what's going on in this PR. Why are we relying on your `@kickstartds/json-schema-to-typescript` fork?

Let's continue the discussion in https://github.com/bcherny/json-schema-to-typescript/issues/551.

PR #560, opened 2023-11-18
Fix issue with `patternProperties` and `strictIndexSignatures` option

### LINKED ISSUE #559: `strictIndexSignatures` option is ignored for single pattern properties
When generating for objects that have `patternProperties` with the `strictIndexSignatures` option, `json-schema-to-typescript` does not honor the `strictIndexSignatures` option.

I traced this down to this area of code:

https://github.com/bcherny/json-schema-to-typescript/blob/c82696300269e6de73818372b07d13a2f9ceeba7/src/parser.ts#L407-L427

Where it returns the `keyName` as `[k: string]` in the `singlePatternProperty` case but doesn't update the `keyName` property of the `ast`, which is used by the `strictIndexSignatures` routine:

https://github.com/bcherny/json-schema-to-typescript/blob/c82696300269e6de73818372b07d13a2f9ceeba7/src/generator.ts#L156-L158

I attempted a fix in this PR, here: https://github.com/bcherny/json-schema-to-typescript/pull/560

## Example schema

```json
{
  "type": "object",
  "properties": {
    "maybe": {
      "type": "string",
    },
    "pattern": {
      "type": "object",
      "properties": {
        "maybe": {
          "type": "string"
        }
      },
      "patternProperties": {
        "leaf|tree": {
          "type": "array"
        }
      }
    }
  }
}
```

## Expected output

```typescript
export interface Experiment {
  maybe?: string;
  pattern?: {
    maybe?: string;
    /**
     * This interface was referenced by `undefined`'s JSON-Schema definition
     * via the `patternProperty` "leaf|tree".
     */
    [k: string]: unknown[] | undefined;
  };
  [k: string]: unknown | undefined;
}
```

## Actual output

```typescript
export interface Experiment {
  maybe?: string;
  pattern?: {
    maybe?: string;
    /**
     * This interface was referenced by `undefined`'s JSON-Schema definition
     * via the `patternProperty` "leaf|tree".
     */
    [k: string]: unknown[];
  };
  [k: string]: unknown | undefined;
}
```

## Related

- https://github.com/bcherny/json-schema-to-typescript/issues/235
- https://github.com/bcherny/json-schema-to-typescript/issues/477

yorinasub17:
Fixes https://github.com/bcherny/json-schema-to-typescript/issues/559

See issue for details.

## NOTE

This is my first time contributing here! I must admit that I do not fully grok the intricacies around the type issues as described in https://github.com/bcherny/json-schema-to-typescript/issues/235 , so I am not 100% certain this is the right fix. That said, this does address my compilation issues when generating from schemas in https://www.schemastore.org/json/.

Also, I am not familiar with `ava` and snapshot testing, so I am not 100% sure I rebuilt the snapshot correctly, so pointers on how to regenerate it would be helpful if I did indeed do something wrong!


bcherny:

This looks like it's already on `master` as of `107dd4284a0b747ffe141949e2f0c429d3616f0d` (#704, "Fix invalid index signature type when patternProperties/additionalProperties combine with named properties", fixes #671) — a broader fix that widens the generated index signature type to also cover sibling named properties, which resolves the same `patternProperties`/`strictIndexSignatures` conflict this PR targets.

I compiled your own test fixture (the `pattern` property from your `options.strictIndexSignatures.ts` addition) against current `master` and it now produces a valid interface with no compile error:

```ts
export interface StrictIndexSignatures {
  maybe?: string;
  complex?: {
    maybe?: string;
    [k: string]: Leaf | string | undefined;
  };
  pattern?: {
    maybe?: string;
    [k: string]: unknown[] | string | undefined;
  };
  [k: string]: string | {...} | {...} | undefined;
}
```

---

Issue #381, opened 2021-04-13
allOf, oneOf dont correctly generate types


italicbold:
The following schema:
```js
        "Test": {
            "type": "object",
            "properties": {
                "prop1": { "type": "string" },
                "prop2": { "type": "string" },
                "prop3": { "type": "string"},
                "prop4": { 
                    "type": "array",
                    "items": {
                        "type": "string"
                    },
                    "minItems": 1
                },
                "prop5": { "type": "string" },
                "prop6": { "type": "string" }
            },
            "allOf": [
                { "required": [ "prop1" ] },
                {
                    "oneOf": [
                        { "required": [ "prop2" ] },
                        { "required": [ "prop3" ] },
                        { "required": [ "prop4" ] },
                        { "required": [ "prop5" ] },
                        { "required": [ "prop6" ] }
                    ]
                }
            ],
            "additionalProperties": false
        }
```

Generates type:

```ts
type Test = {
    [k: string]: unknown;
} & (
    | {
          [k: string]: unknown;
      }
    | {
          [k: string]: unknown;
      }
    | {
          [k: string]: unknown;
      }
    | {
          [k: string]: unknown;
      }
    | {
          [k: string]: unknown;
      }
);
```

Expected type:
```ts
type Test = {
    prop1: string;
} & (
    { prop2: string } |
    { prop3: string } |
    { prop4: string[] } |
    { prop5: string } |
    { prop6: string }
);
```

italicbold:
Similar to https://github.com/bcherny/json-schema-to-typescript/issues/378 ?

Used version 10.1.4

RickZack:
@italicbold the mentioned issue is related to the use of allOf, anyOf with properties, and in that case there was an error in the provided example. Here the problem is related to in processing `requested` property when inside oneOf, anyOf. I think it is different.

john-twigg-ck:
I'm seeing the same problem. Usage of `anyOf` causes schema to devolve to `[k: string]: unknown;`

FloEdelmann:
I found that nesting `oneOf`/`anyOf`/`allOf` causes the issue:

**Minimum example schema:**

```json5
{
  "oneOf": [
    {
      "properties": {
        "speed": { "type": "string" }
      },
      "oneOf": [ // or "allOf" or "anyOf"
        { "required": ["speed"] }
      ]
    }
  ]
}
```

<details>
<summary>Example schema with a bit more context</summary>

```json
{
  "oneOf": [
[...]

PR #566, opened 2024-01-03
Fix parsing of schemas that are combinations of unnamed schemas and oneOf/allOf/...

anzev:
This is an attempt to fix the issue described in my comment [here](https://github.com/bcherny/json-schema-to-typescript/issues/381#issuecomment-1854154150).

Basically, the issue arises when parsing a schema that is a combination of a `oneOf`/`allOf`/... and an unnamed schema (see example below), for example, the output interface for the array items would completely ignore the `properties` part of the schema.

This was due to the way how schema type matching was done. It would match only as `ONE_OF` and not a combination of `[ONE_OF, UNNAMED_SCHEMA]`. If we would add an `$id` to the definition, the schema would be parsed correctly as `[ONE_OF, NAMED_SCHEMA]`.

This PR attempts to fix this issue by changing how `UNNAMED_SCHEMA` is matched by adding an explicit condition. So far it was handled as: if nothing else matches, default to `UNNAMED_SCHEMA`, which is why it never matched in scenarios described above.

I've added two tests and updated several existing tests, since the new condition generates types that were previously missed.

Please check it out and let me know any improvement and if this is a valid solution in your opinion.

Thanks!

-----
Example:

```json
{
  "type": "object",
  "$schema": "http://json-schema.org/draft-07/schema",
  "properties": {
    "things": {
      "type": "array",
      "minLength": 1,
      "description": "A list of things",
      "items": {
        "type": "object",
        "required": ["thingProp1", "thingProp2"],
        "oneOf": [
          {
            "type": "object",
            "properties": {
              "propOption1": {
                "type": "string"
              }
            },
            "required": ["propOption1"]
          },
          {
            "type": "object",
            "properties": {
              "propOption2": {
                "type": "number"
              }
            },
            "required": ["propOption2"]
          }
        ],
        "properties": {
          "thingProp1": {
            "type": "string"
          },
          "thingProp2": {
            "type": "string"
          }
        }
      }
    }
  }
}
```


joeferner:
This PR works great for me.

I tested this [PR](https://github.com/bcherny/json-schema-to-typescript/pull/571) without success.

bcherny:

This looks like it's already on `master` as of `d8618e5648c473bbf580a709684394d3dc2de523` (#708, "Fix properties dropped when combined with oneOf/anyOf/allOf on a nested schema", fixes #630) — which fixes the exact `typesOfSchema`/`UNNAMED_SCHEMA` matching gap this PR describes, for nested (non-`$id`) schemas.

I compiled the exact example schema from your PR description against current `master` and `thingProp1`/`thingProp2` are correctly intersected in, no longer dropped:

```ts
export interface Test {
  things?: ((
    | { propOption1: string; [k: string]: unknown }
    | { propOption2: number; [k: string]: unknown }
  ) & {
    thingProp1: string;
    thingProp2: string;
    [k: string]: unknown;
  })[];
  [k: string]: unknown;
}
```

---

PR #573, opened 2024-01-25
add `commentStyle` option

sudhanshug16:
Adds a `commentStyle` option. Allows the user to choose from block or line comments.

I am using this package with LLMs. Since machines don't really care about the nicely formatted TSdoc comments, I wanted to switch to line comments `// this is a comment`. This way I can save on tokens and keep the prompt shorter to reduce hallucinations.

This is a non-breaking change as it keeps the default as `block`.

### DIFF
--- a/src/generator.ts
+++ b/src/generator.ts
@@ -304,7 +304,7 @@ function generateInterface(ast: TInterface, options: Options): string {
       )
       .map(
         ([isRequired, keyName, ast, type]) =>
-          (hasComment(ast) && !ast.standaloneName ? generateComment(ast.comment, ast.deprecated) + '\n' : '') +
+          (hasComment(ast) && !ast.standaloneName ? generateComment(ast.comment, ast.deprecated, options) + '\n' : '') +
           escapeKeyName(keyName) +
           (isRequired ? '' : '?') +
           ': ' +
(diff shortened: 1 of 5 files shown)


bcherny:
Thanks for the contribution! An e2e test would be great. Here's an example: https://github.com/bcherny/json-schema-to-typescript/blob/master/test/e2e/options.unreachableDefinitions.ts.

src/generator.ts
```diff
@@ -316,21 +316,30 @@ function generateInterface(ast: TInterface, options: Options): string {
...
+    commentLines.push(...comment.split('\n').map(_ => linePrefix + _))
   }
-  commentLines.push(' */')
-  return commentLines.join('\n')
+
+  if (options.commentStyle === 'block') {
+    commentLines.unshift('/**')
+    commentLines.push(' */')
+
+    return commentLines.join('\n')
```
bcherny, line 334: Nit

```ts
return [
  '/**',
  ...commentLines,
  ' */'
].join('\n')
```

src/generator.ts
```diff
@@ -316,21 +316,30 @@ function generateInterface(ast: TInterface, options: Options): string {
...
+  }
+
+  return '// ' + commentLines.map(_ => (_.endsWith('.') ? _ : _ + '.')).join(' ')
```
bcherny, line 337: Appending `.` seems like a stylistic choice better left to consumers.

PR #575, opened 2024-02-01
Added option to remove comment about interface referencer

Stmated:
When generating code using `unreachableDefinitions` it adds a comment to the interface stating `This interface was referenced by...`.
I find this quite annoying. So I added an option to not include those comments.


bcherny (changes requested):
Thanks for the contribution, but I'm not sure this change is worth the added complexity it introduces. We can revisit if there is significant interest from others (please feel welcome to open an issue).

Stmated:
> Thanks for the contribution, but I'm not sure this change is worth the added complexity it introduces. We can revisit if there is significant interest from others (please feel welcome to open an issue).

You know what would bring even less complexity? Just removing the comment without any setting of enabling it.

PR #576, opened 2024-02-06
Make schema titles as property types optional

ethanpmullen:
By default, this library hoists the "title" filed of a property as the name of the property type. For example, the following schema:
```
export const input = {
  title: 'Example Schema',
  type: 'object',
  properties: {
    firstName: {
      title: 'First Name',
      type: 'string',
    },
    lastName: {
      title: 'Last Name',
      id: 'lastName',
      type: 'string',
    }
  },
  required: ['firstName', 'lastName'],
}
```
Produces this type:
```
export type FirstName = string;

export type LastName = string;

      firstName: FirstName;
      lastName: LastName;
      [k: string]: unknown;
}
```

Per the [JSON schema specifications](https://json-schema.org/understanding-json-schema/reference/annotations), titles are annotations for description purposes, so it can often be a clearer interface to use the actual type (like `string` or `boolean`) rather than the "Title" in the JSON schema file. 

To use this feature, all you have to do is pass in `useSchemaTitleAsPropertyType=false`, and the following schema will be produced:
```
      firstName: string;
      lastName: string;
      [k: string]: unknown;
}
```
Default behavior remains intact and tests have been added to support this behavior.


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "json-schema-to-typescript",
-  "version": "13.1.2",
+  "name": "@linktr.ee/json-schema-to-typescript",
```
bcherny, line 2: Please remove this change from this PR.

src/index.ts
```diff
@@ -179,7 +193,15 @@ export async function compile(schema: JSONSchema4, name: string, options: Partia
   const formatted = format(generated, _options)
   log('white', 'formatter', time(), '✅ Result:', formatted)
 
-  return formatted
+  return {
+    typescript: formatted,
```
bcherny, line 197: This is a breaking change. Please avoid changing the result shape, and instead focus on making the most targeted change possible.

src/index.ts
```diff
@@ -76,6 +76,10 @@ export interface Options {
    * Generate unknown type instead of any
    */
   unknownAny: boolean
+  /**
+   * Expand the title field of each property into a separate type
+   */
+  useSchemaTitleAsPropertyType: boolean
```
bcherny, line 82: Let's call this `inferTypeAliasFromTitle`

bcherny (changes requested):
This change seems reasonable overall, but the implementation needs a bit more iteration. Could you please:

1. Implement this as a normalizer step instead, moving the `title` to a `description` when  `options.inferTypeAliasFromTitle` is `false` (thereby, isolating the change from all downstream code).
2. Add a test for the normalizer step, in addition to the e2e tests you added.
3. Ensure that titles are emitted as code comments. I don't think we should elide them from the output entirely.

PR #585, opened 2024-03-27
feat: support `customName` options

### LINKED ISSUE #181: Add option to control type naming.

antfu:
Redo #215
Fix #181

Different from #215, this PR does not include CLI support - since the options accept a function, I suppose it would be more useful for programmatic usages.


src/index.ts
```diff
@@ -76,6 +77,10 @@ export interface Options {
    * Generate unknown type instead of any
    */
   unknownAny: boolean
+  /**
+   * Custom function to provide a type name for a given schema
+   */
+  customName?: (schema: LinkedJSONSchema, keyNameFromDefinition: string | undefined) => string
```
antfu, line 83: A bit different from the proposal in https://github.com/bcherny/json-schema-to-typescript/issues/181#issuecomment-421885339, I think it would be more useful to expose the schema object directly, as both `id` and `title` comes from the schema object. This would allow demands like #456 to be easily achieved from the userland.

src/parser.ts
```diff
@@ -343,7 +343,11 @@ function standaloneName(
   schema: LinkedJSONSchema,
   keyNameFromDefinition: string | undefined,
   usedNames: UsedNames,
+  options: Options,
 ): string | undefined {
+  if (options.customName) {
+    return options.customName(schema, keyNameFromDefinition)
```
bcherny, line 349: I wonder if it would be more useful to make it so `customName` can return `string | null`, then if the result is `null`, we fall back to the logic below. That would make it so you can customize some names, but you don't have to handle very single name.

Would that be better/worse for your use case, for example?

antfu: I would pretty much prefer to allow nullish returns to fallback. I didn't do it because I saw you mention it here: https://github.com/bcherny/json-schema-to-typescript/pull/215#discussion_r261842905

If you change your mind, I am happy to push an update

bcherny: Returning `string | void` is fine -- I disagree with old-me :)

bcherny (changes requested):
See comment

bcherny (approved):
Thanks for the contribution!

Issue #577, opened 2024-02-07
Add yaml support


codan84:
If one writes their schemas in `yaml` file format, is there a simple way to generate types using the CLI tool?
If not - could a support for that be added please?

alexanderkrum:
@bcherny picking this up

alexanderkrum:
https://github.com/bcherny/json-schema-to-typescript/pull/588

bcherny:
Published 14.0.2.

PR #588, opened 2024-04-10
feat: [577] Add yml support

alexanderkrum:
This PR is adding support for the yml files.

Fixes: #577


src/index.ts
```diff
@@ -107,19 +108,36 @@ export const DEFAULT_OPTIONS: Options = {
...
-    () => {
-      throw new TypeError(`Error parsing JSON in file "${filename}"`)
-    },
-  )
+
+  let schema: JSONSchema4
+
+  if (isYml(filename)) {
+    schema = Try<JSONSchema4>(
+      () => JSON.parse(JSON.stringify(yaml.load(contents.toString()))),
```
bcherny, line 127: Can you remove the `JSON.parse(JSON.stringify())`?

src/index.ts
```diff
@@ -107,19 +108,36 @@ export const DEFAULT_OPTIONS: Options = {
   unknownAny: true,
 }
 
+function isYml(filename: string) {
+  return filename.split('.').pop() === 'yml'
```
bcherny, line 112: 1. Please use `.endsWith`
2. We should support `.yaml` too

```ts
return filename.endsWith('.yaml') || filename.endsWith('.yml')
```

README.md
```diff
@@ -4,7 +4,7 @@
 [npm]: https://img.shields.io/npm/v/json-schema-to-typescript.svg?style=flat-square
 [mit]: https://img.shields.io/npm/l/json-schema-to-typescript.svg?style=flat-square
 
-> Compile json schema to typescript typings
+> Compile json/yml schema to typescript typings
```
bcherny, line 7:
```suggestion
> Compile json/yaml schema to typescript typings
```

bcherny (changes requested):
Thanks for the contribution! Let's take one more pass at it, then land it.

G-Rath:
While I understand why it would be cool for this to be supported without any extra steps, it will mean that _all_ users have to pull in extra dependencies for something that they might not be using.

For a library this can be easily solved by making the dependency an optional peer, but for CLIs that's not as easy.

There's also the argument of "where does it stop?" What would jsonc, json5, and toml support?

I wonder if it would be worth seeing if it might be better to encourage (and support, if not already) passing content in via pipes

I.e `js-yaml` provides a CLI for converting to json, so I wouldn't be surprised if right now you can do something like `npx js-yaml path/to/schema.yml | npx json-schema-to-typescript -`

bcherny:
@G-Rath That would be cleaner for sure. But because JSON Schemas can reference other schemas, and because we already support .yaml-backed references (https://github.com/APIDevTools/json-schema-ref-parser/tree/main/lib/parsers), I see supporting .yaml for root schemas as more about completeness. In theory if there was a way to make .yaml parsing/pre-processing pluggable all the way through, that would be nicer for sure!

src/index.ts
alexanderkrum: Updated as suggested

src/index.ts
alexanderkrum: Updated as suggested

G-Rath:
@bcherny fwiw while I've never used it like this myself, `npx` and co let you pass multiple packages which get installed in the same temp context so my understanding is you should be able to do something like `npx -p js-yaml -p json-schema-to-typescript -- json-schema-to-typescript path/to/schema.yml`

However, since `js-yaml` is already a dependency of `json-schema-ref-parser`, using it directly seems fine to me

bcherny:
Looks good! Please rebase to fix the merge conflict, then I'll merge this.

bcherny (approved):

bcherny:
Thanks for the contribution! Will include this in the next release, next week.

PR #591, opened 2024-04-16
fix: force resolution of es5-ext

yasinkocak:
Force resolution of es5-ext is being flagged as malware.
Because of postinstall script of a dependency printing out a message to users in Russian time zones. https://github.com/medikoo/es5-ext/issues/116
Relevant to https://github.com/medikoo/es5-ext/issues/186 ("package being detected as a virus")

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,9 @@
     "watch": "tsc -w",
     "watch:test": "ava -w"
   },
+  "resolutions": {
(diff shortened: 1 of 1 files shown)


bcherny:
Feel free to do this in your project's package.json, but I don't think we should do this here. Thanks for the contribution!

PR #596, opened 2024-05-25
Update README.md: fix definition of customName

archiloque:
Added escape chars to avoid the `|` in type definition being interpreted as cell separator


bcherny (approved):
Thanks for the fix!

Issue #602, opened 2024-06-23
Generation for intersection/named schemas is order-dependent


altearius:
- Given a schema that contains a named definition (`Level2B`),
- And that named definition is referenced in multiple locations,
- And that named schema is also an intersection type (`allOf` in this example),

Then when parsed, the generated TypeScript will contain the correct reference only for the _first_ location in which the named schema is encountered, during a depth-first traversal.

Subsequent references to the same schema will be generated as though they were only the intersection type, and not the named schema.

# Example

Given the following schema:

```yaml
description: Top Level
type: object
oneOf:
  - $ref: '#/definitions/Level2A'
  - $ref: '#/definitions/Level2B'

definitions:
  Level2A:
    description: Level2A
    type: object
    allOf: [$ref: '#/definitions/Base']
    properties:
      level_2A_ref: { $ref: '#/definitions/Level2B' }
  Level2B:
    description: Level2B
    type: object
    allOf: [$ref: '#/definitions/Base']
    properties:
      level_2B_prop: { const: xyzzy }
  Base:
    description: Base
    type: object
    properties:
      base_prop: { type: string }
```

The current resulting TypeScript will be (comments adjusted for clarity):

```ts
// Incorrect: should be `type Demo = Level2A | Level2B`
// Note that the Level2B type at this location is the _second_ instance to
// Level2B during a depth-first traversal.
export type Demo = Level2A | Level2B1;

export type Level2A = Level2A1 & {
  // Correct in this location because this property is reached first in a
  // depth-first traversal.
  level_2A_ref?: Level2B;
  [k: string]: unknown;
};

export type Level2A1 = Base;

export type Level2B = Level2B1 & {
  level_2B_prop?: "xyzzy";
  [k: string]: unknown;
};

export type Level2B1 = Base;

export interface Base {
  base_prop?: string;
  [k: string]: unknown;
}
```

# Root Cause

In `parser.ts`, [lines 57 - 75][1], when schema that matches multiple "types" is encountered, the parser generates a new `ALL_OF` intersection schema to contain each sub-type, then adds each sub-type to the new `ALL_OF` schema.

Each sub-type is then parsed sequentially. During this process, `maybeStripNameHints` is called, which mutates the schema by removing the `$id`, `description`, and `name` properties.

Notably, these properties are used by `typesOfSchema` to detect the `NAMED_SCHEMA` type. As a result, this schema object will never again be detected as a `NAMED_SCHEMA` type.

[...]

PR #603, opened 2024-06-23
Intersection schema generation is order-dependent

altearius:
Addresses #602 and #597


src/applySchemaTyping.ts
```diff
@@ -0,0 +1,20 @@
+import type {LinkedJSONSchema} from './types/JSONSchema'
+import {typesOfSchema} from './typesOfSchema'
+
+export function applySchemaTyping(schema: LinkedJSONSchema) {
+  const types = typesOfSchema(schema)
+  schema.$types = types
+
+  if (types.length > 1) {
+    schema.$intersection = {
```
bcherny, line 9: `$intersection` and `$types` may conflict with schema properties (in JSON-Schema, users can define custom properties). Instead, can you use Symbols (similar to `Parent`)? This name collision could also be good to add as a test case.

https://github.com/bcherny/json-schema-to-typescript/blob/31993def993b610ba238d3024260129e31ddc371/src/types/JSONSchema.ts#L43

src/applySchemaTyping.ts
```diff
@@ -0,0 +1,20 @@
...
+      $id: schema.$id,
+      allOf: [],
+      description: schema.description,
+      title: schema.title,
+    }
+
+    delete schema.$id
+    delete schema.description
+    delete schema.name
```
bcherny, line 18: What about `title`?

src/parser.ts
```diff
@@ -54,13 +55,13 @@ export function parse(
 
   // Be careful to first process the intersection before processing its params,
   // so that it gets first pick for standalone name.
+  const intersectionSchema = schema.$intersection
+  if (!intersectionSchema) {
+    throw new Error('Expected intersection schema')
```
bcherny, line 60:
```suggestion
    throw new ReferenceError('Expected intersection schema')
```

src/types/JSONSchema.ts
```diff
@@ -71,6 +72,8 @@ export interface LinkedJSONSchema extends JSONSchema {
 }
 
 export interface NormalizedJSONSchema extends LinkedJSONSchema {
+  $intersection?: NormalizedJSONSchema
+  $types: ReturnType<typeof typesOfSchema>
```
bcherny, line 76: Nit: please type this explicitly to keep it simple.
```suggestion
  $types: readonly [SchemaType, ...SchemaType[]]
```

bcherny (changes requested):
Thank you for the contribution!

Your approach looks good to me. I tried a more surgical fix in https://github.com/bcherny/json-schema-to-typescript/pull/605/files, but like that your approach is more generic and cleaner.

A few minor things before we merge this:

1. Would you mind splitting your PR into two commits: one to add the new test cases, and one to make code changes and change the output of your test cases? That would make it more clear how your changes change the output of the test cases.
2. Would you mind simplifying the naming in your cases (eg. `A` instead of `Level2A`)? (example: https://github.com/bcherny/json-schema-to-typescript/pull/605/commits/1252285c9f934aff7e1875965e779866bab228c4)

src/applySchemaTyping.ts
altearius: This suggestion significantly improved the code quality, thank you.

src/applySchemaTyping.ts
altearius: I missed `title` as it was not handled by the original [`maybeStripNameHints` function][1], though I have now added it.

[1]: https://github.com/bcherny/json-schema-to-typescript/blob/31993def993b610ba238d3024260129e31ddc371/src/utils.ts#L341-L352 'maybeStripNameHints'

altearius:
Thank you for the code review! This is ready for a second pass.

test/__snapshots__/test/test.ts.md
```diff
@@ -1861,6 +1861,92 @@ Generated by [AVA](https://avajs.dev).
...
+
+    `/* eslint-disable */␊
+    /**␊
+     * This file was automatically generated by json-schema-to-typescript.␊
+     * DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,␊
+     * and run json-schema-to-typescript to regenerate this file.␊
+     */␊
+    ␊
+    export type Intersection = A | B;␊
+    export type A = A1 & {␊
```
bcherny, line 1876: I think I missed this earlier -- this is a case where the output in #605 is better than #603. I wonder if there's a way to improve this output, so we don't generate `A1` and `B1`, and instead just use `Base` directly.

Like I did here: https://github.com/bcherny/json-schema-to-typescript/pull/605/files#diff-34615f003eb49f90c3044b8fafebe3fe6b162e290b9b724082a6d72f58cfa4afR1876

test/__snapshots__/test/test.ts.md
```diff
@@ -1861,6 +1861,92 @@ Generated by [AVA](https://avajs.dev).
...
+
+    `/* eslint-disable */␊
+    /**␊
+     * This file was automatically generated by json-schema-to-typescript.␊
+     * DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,␊
+     * and run json-schema-to-typescript to regenerate this file.␊
+     */␊
+    ␊
+    export type Intersection = Car | Truck;␊
+    export type Car = Car1 & {␊
```
bcherny, line 1928: And here. This output is cleaner, if you can achieve something similar with your approach: https://github.com/bcherny/json-schema-to-typescript/pull/605/files#diff-34615f003eb49f90c3044b8fafebe3fe6b162e290b9b724082a6d72f58cfa4afR1927

test/__snapshots__/test/test.ts.md
```diff
@@ -448968,8 +448968,7 @@ Generated by [AVA](https://avajs.dev).
      * and run json-schema-to-typescript to regenerate this file.␊
      */␊
     ␊
-    export type UnionWithProps = UnionWithProps1 & UnionWithProps2;␊
-    export type UnionWithProps1 =␊
+    export type UnionWithProps = (␊
```
bcherny, line 448971: Nice

bcherny (changes requested):
Looking better!

The output is a nice improvement across a number of tests. I wonder if there might be a way to improve the output more for the test cases you added.

To help debug, try adding `export const only = true` to your test case, then running `VERBOSE=1 npm test` to enable debug tracing.

test/__snapshots__/test/test.ts.md
altearius: Thanks! I took a try at that in 7d3f495 by pulling schema's pre-existing `allOf` rule into the intersection. Would you mind taking a third pass and let me know if I missed anything?

test/__snapshots__/test/test.ts.md
```diff
@@ -447961,15 +447871,78 @@ Generated by [AVA](https://avajs.dev).
             nohoist?: string[];␊
             [k: string]: unknown;␊
           };␊
-      jspm?: JSONSchemaForNPMPackageJsonFiles1;␊
+      jspm?: JSONSchemaForNPMPackageJsonFiles;␊
       /**␊
        * Any property starting with _ is valid.␊
        *␊
-       * This interface was referenced by \`JSONSchemaForNPMPackageJsonFiles2\`'s JSON-Schema definition␊
+       * This interface was referenced by \`undefined\`'s JSON-Schema definition␊
```
bcherny, line 447878: This seems to have regressed

test/__snapshots__/test/test.ts.md
```diff
@@ -225147,7 +225124,7 @@ Generated by [AVA](https://avajs.dev).
     /**␊
      * The list of custom alert time-window rules.␊
      */␊
-    timeWindowRules?: ((ActiveConnectionsNotInAllowedRange | AmqpC2DMessagesNotInAllowedRange | MqttC2DMessagesNotInAllowedRange | HttpC2DMessagesNotInAllowedRange | AmqpC2DRejectedMessagesNotInAllowedRange | MqttC2DRejectedMessagesNotInAllowedRange | HttpC2DRejectedMessagesNotInAllowedRange | AmqpD2CMessagesNotInAllowedRange | MqttD2CMessagesNotInAllowedRange | HttpD2CMessagesNotInAllowedRange | DirectMethodInvokesNotInAllowedRange | FailedLocalLoginsNotInAllowedRange | FileUploadsNotInAllowedRange | QueuePurgesNotInAllowedRange | TwinUpdatesNotInAllowedRange | UnauthorizedOperationsNotInAllowedRange)[] | string)␊
+    timeWindowRules?: (TimeWindowCustomAlertRule[] | string)␊
```
bcherny, line 225127: This looks like a regression. Asked about it here to understand the expected behavior: https://github.com/json-schema-org/json-schema-spec/issues/1529

test/__snapshots__/test/test.ts.md
altearius: Just a heads up: I haven't lost interest in moving this forward. I've had some deadlines come up, but I plan to return to this over the weekend.

altearius: Hello! Thank you for your patience. I've taken a look at this regression and I'm not sure what the expected behavior should be. Could you advise?

This is what I observe, simplified to the relevant bits.

Given this schema:

```yaml
title: JSON schema for NPM package.json files
type: object
patternProperties:
  '^_':
    description: Any property starting with _ is valid.
properties:
  # etc, etc, etc, actual properties clipped
```

The original code produced this structure:

```ts
export type JSONSchemaForNPMPackageJsonFiles =
  JSONSchemaForNPMPackageJsonFiles1 & JSONSchemaForNPMPackageJsonFiles2;

export type JSONSchemaForNPMPackageJsonFiles1 = { [k: string]: unknown; };

export interface JSONSchemaForNPMPackageJsonFiles2 {␊
  /**
   * Any property starting with _ is valid.
   * 
   * This interface was referenced by \`JSONSchemaForNPMPackageJsonFiles2\`'s
   * JSON-Schema definition via the \`patternProperty\` "^_".
   */
  [k: string]: any;
}
```

This branch produces a different structure:

```ts
export type JSONSchemaForNPMPackageJsonFiles =
  { [k: string]: unknown; } &
  {
    /**
     * Any property starting with _ is valid.
     * 
     * This interface was referenced by \`undefined\`'s JSON-Schema definition
     * via the \`patternProperty\` "^_".
     */
    [k: string]: any;
  };
```

While I agree that the change in the comment from `JSONSchemaForNPMPackageJsonFiles2` to `undefined` is unfortunate, this behavior is in line with other instances in which this comment is generated due to the presence of `patternProperties` on the root schema.

For instance, here is the current output of the `intersection.2.ts` test on the `master` branch, which has a similar scenario:

```ts
export type Intersection = (A & B) & {
  c: string;
  d: string;
  /**
   * This interface was referenced by \`undefined\`'s JSON-Schema definition
   * via the \`patternProperty\` "^x-".
   */
  [k: string]: unknown;
};

export interface A {
  a?: string;
  [k: string]: unknown;
}

export interface B {
  b?: string;
  [k: string]: unknown;
}
```

I would argue that neither behavior is ideal, and an improvement might be to suppress the comment entirely if the interface is not named.

I can make this change, but I worry that it may not be what you have in mind, and that it feels like something that should be a separate PR.

Could you please provide direction on how this might be resolved?

test/__snapshots__/test/test.ts.md
altearius: Given this schema:

```yaml
type: object
required: [ruleType, timeWindowSize]

oneOf:
  - $ref: '#/definitions/ActiveConnectionsNotInAllowedRange'
  # ...etc...
  - $ref: '#/definitions/UnauthorizedOperationsNotInAllowedRange'

properties:
  ruleType:
    type: string
    enum: [ TimeWindowCustomAlertRule ]

  timeWindowSize:
    type: string
```

The old behavior is:

```ts
export interface DeviceSecurityGroupProperties {
  // Note: does **not** include the ruleType or timeWindowSize properties
  timeWindowRules?:
    | (
        | ActiveConnectionsNotInAllowedRange
        // ...etc...
        | UnauthorizedOperationsNotInAllowedRange
      )[]
    | string;
  [k: string]: unknown;
}

export type TimeWindowCustomAlertRule = {
  ruleType: "TimeWindowCustomAlertRule";
  timeWindowSize: string;
  [k: string]: unknown;
} & (
  | ActiveConnectionsNotInAllowedRange
  // ...etc...
  | UnauthorizedOperationsNotInAllowedRange
);
```

The new behavior is:

```ts
export interface DeviceSecurityGroupProperties {
  // Note: Now includes ruleType and timeWindowSize.
  timeWindowRules?: TimeWindowCustomAlertRule[] | string;
}

export type TimeWindowCustomAlertRule = {
  ruleType: "TimeWindowCustomAlertRule";
  timeWindowSize: string;
  [k: string]: unknown;
} & (
  | ActiveConnectionsNotInAllowedRange
  // ...etc...
  | UnauthorizedOperationsNotInAllowedRange
);
```

I would argue that the new behavior is more nearly correct. My reasoning is based on the definition of the `oneOf` keyword in the [JSON Schema spec][1]:

> 10.2.1.3. oneOf
>
> This keyword's value MUST be a non-empty array. Each item of the array MUST be a valid JSON Schema.
>
> An instance validates successfully against this keyword if it validates successfully against exactly one schema defined by this keyword's value.

This definition does not address the possibility that `oneOf` may be combined with additional keywords, such as `properties`. However, one reasonable interpretation is that the `properties` keyword should be applied in addition to the `oneOf` keyword. In such a case, both `oneOf` and `properties` (or other keywords) would all apply.

This interpretation aligns with [section 10.1][3]:

> 10.1 Keyword Independence
>
> Schema keywords typically operate independently, without affecting each other's outcomes.

This suggests that `oneOf` and `properties` should _both_ affect the output.

This interpretation is also in line with the [Factoring Schemas][2] example, which demonstrates how one schema:

```yaml
oneOf:
  - { type: number, multipleOf: 5 }
  - { type: number, multipleOf: 3 }
```

Can be factored into a schema that combines `oneOf` with additional keywords (`type`, in this case, but I presume an example with `properties` would be equally valid):

```yaml
type: number
oneOf:
  - { multipleOf: 5 }
  - { multipleOf: 3 }
```

I realize that other interpretations of the `oneOf` schema are possible, but this is what feels most intuitive to me.  Please let me know if other changes need to be made.

[1]: https://json-schema.org/draft/2020-12/json-schema-core#section-10.2.1.3 'oneOf'
[2]: https://json-schema.org/understanding-json-schema/reference/combining#factoringschemas 'Factoring Schemas'
[3]: https://json-schema.org/draft/2020-12/json-schema-core#name-keyword-independence 'Keyword Independence'

test/__snapshots__/test/test.ts.md
bcherny: That seems reasonable -- we can improve this as a follow-up.

test/__snapshots__/test/test.ts.md
bcherny: I think that's a reasonable interpretation, but as you said, there may be other interpretations. Implicit intersection schemas like this make me scratch my head sometimes. I *think* your interpretation is correct, and the new behavior is correct, but since it's a subtle breaking change that will affect the largest consumers of JSTT, I want to double check with JSON-Schema maintainers first.

I bumped the Github issue to see is someone is willing to clarify.

bcherny:
Published as v15.0.0.

PR #615, opened 2024-08-09
Add ability to override protocol from http:// to file://  in schema ids

theanurin:
TDB


bcherny:
Without a summary/issue it's hard to tell what the motivation for this PR is, but from reading your code, I think you can pass in custom JSON Schema Ref Parser options to do what you need. See https://apitools.dev/json-schema-ref-parser/docs/options.html

PR #616, opened 2024-08-12
fix: fix bad validation of definitions

aloisklink:
Sometimes, `traverse` accidentally traverses a non-schema object, e.g. instead traversing the `properties` object of a schema.

This then causes validation errors if one of these properties is "invalid" (e.g. you have a key called `deprecated`).

Using `traverse` instead of `traverseObjectKeys` seems to fix this, since then we don't traverse on objects that contain `BLACKLISTED_KEYS`.

For an example, see the test case I added. On the current `master` branch, shows the following error:

```
[
  'Error at key "properties" in file "deprecated": deprecated must be a boolean'
]
```

For a real world use-case, I was trying to use the [OpenAPI 3.1 JSON Schema](https://github.com/OAI/OpenAPI-Specification/blob/08f92adcade2b3d584e8039ae056a9bf82fb6424/schemas/v3.1/schema.json) in my app and I got this validation error.

[^1]: Unfortunately, even after fixing this bug, it looks like OpenAPI 3.1's JSON Schema uses some new features in JSON Schema that the `@apidevtools/json-schema-ref-parser` library does not yet support: https://github.com/APIDevTools/json-schema-ref-parser/issues/145


bcherny:

I built `master` at `d8618e5` and transcribed the repro from your description (a schema with a `properties.deprecated` sub-schema, i.e. a property literally named `deprecated`):

```js
const {compile} = require('json-schema-to-typescript')

const input = {
  title: 'deprecated',
  type: 'object',
  properties: {
    deprecated: {
      type: 'object',
      properties: {foo: {type: 'string'}},
    },
  },
}

await compile(input, 'Deprecated', {$refOptions: {resolve: {file: false, http: false}}})
// => compiles cleanly, no "deprecated must be a boolean" ValidationError
```

This compiles cleanly on current `master` — no `ValidationError` is thrown. `traverse` in `src/utils.ts` no longer confuses a schema's own `properties.deprecated` sub-schema with its `deprecated` keyword, so the validator's "deprecated must be a boolean" rule isn't misapplied here anymore.

Closing as no longer reproducing — but if you're still seeing this on latest, say so and I'll reopen.



---

PR #617, opened 2024-08-17
Remove unused dependency

benmccann:
Also fixes two deprecation warnings:
```
npm warn deprecated @types/rimraf@4.0.5: This is a stub types definition. rimraf provides its own type definitions, so you do not need this installed.
npm warn deprecated @types/mkdirp@2.0.0: This is a stub types definition. mkdirp provides its own type definitions, so you do not need this installed.
```


bcherny:
Thanks!

PR #620, opened 2024-08-20
Use tinyglobby for globbing

benmccann:
https://npmgraph.js.org/?q=glob - 26 dependencies
https://npmgraph.js.org/?q=tinyglobby - 2 dependencies


benmccann:
hmm. this will need some more work

PR #622, opened 2024-08-20
Use tinyglobby for globbing

benmccann:
the snapshot was updated because `tinyglobby` returns the files in a different order than `glob`

https://npmgraph.js.org/?q=glob - 26 dependencies
https://npmgraph.js.org/?q=tinyglobby - 2 dependencies


bcherny:
I appreciate the PR, but we won't be going forward with it. `glob` has 140M downloads/week on NPM, vs. `tinyglobby`'s 140k. Let's use the boring option unless there is a strong reason not to.

bcherny:
What problem are you trying to solve? Is there a specific issue you're running into? The context could be helpful.

benmccann:
I'm trying to reduce the number of dependencies that need to be installed when installing payload CMS. Almost half of the CMS's dependencies come from this library, but a majority of those really aren't necessary for the core functionality provided by this library

bcherny:
Gotcha. I just published v15.0.1, which removes 41 dependencies (39% of JSTT's dependencies). Hopefully that helps.

Moving the CLI to a new package would be a reasonable as a way to reduce the size of JSTT's dependency tree, but it will have a cost for consumers -- they will now need to download a separate package, fiddle with keeping the two up to date and in sync, and so on.

If you want to reduce the size of the dependency tree even more, you're more than welcome to open an issue. If there is significant interest from others, I'd be open to making the change.

I know this isn't the outcome you were hoping for, but I hope it's a reasonable compromise. You're also welcome to put up PRs for any low-hanging fruit I missed, along the lines of c09dbcddddadabb0c041cc9ab7f816576ae9430c, a1d22d89708cc1c5a49d6b1325068ccaf1144a5d, and 108f144b379b7ce3b9c7ec30dfdf640504c0c3b7.

benmccann:
That is great. Thank you! I really appreciate your efforts on this!

> If you want to reduce the size of the dependency tree even more, you're more than welcome to open an issue. If there is significant interest from others, I'd be open to making the change.

Done! I've also proposed another alternative: https://github.com/bcherny/json-schema-to-typescript/issues/623

> https://github.com/bcherny/json-schema-to-typescript/commit/c09dbcddddadabb0c041cc9ab7f816576ae9430c

Thank you! I had some thoughts on removing `cli-color` as well. Another way to do it would be to replace it with `picocolors`, which is quite popular (55m downloads / week). There's a slight blocker, which is that this library uses bright colors and those were only recently added to picocolors and not yet released, but we could consider it in another week or two perhaps

> Moving the CLI to a new package would be a reasonable as a way to reduce the size of JSTT's dependency tree, but it will have a cost for consumers -- they will now need to download a separate package, fiddle with keeping the two up to date and in sync, and so on.

I don't think it necessarily would, but we can take the discussion to the issue where I've provided some more details: https://github.com/bcherny/json-schema-to-typescript/issues/623

PR #628, opened 2024-09-03
switch to picocolors for console formatting

benmccann:
https://npmgraph.js.org/?q=cli-color - 14 dependencies
https://npmgraph.js.org/?q=picocolors - 0 dependencies


bcherny:
Thanks for the contribution!

Is there a benefit of this approach for your use case? Presumably, this would *add* a dependency to your project.

benmccann:
It doesn't require someone to install an undocumented dependency in order to get highlighting. We could keep it as being included at runtime, but it might be nice to mention that you need to install it and this would still be a lightweight package that users are more likely to already have in their dependency tree

bcherny:
Log highlighting is something you can only enable locally, in development -- I'm not sure it would make much sense in production, unless you have a use case in mind?

benmccann:
Yeah, I agree it doesn't really make sense in production. I've moved it back to dynamically requiring the library

JounQin:
Any news?

`cli-color` -> `es5-ext` detected as a virus

https://github.com/medikoo/es5-ext/issues/186

benmccann:
To be clear, there's no security concern. The package prints a message instructing users how to circumvent Russian censorship regarding the war in Ukraine using Tor and a Russian anti-virus company is flagging it as a result.

JounQin:
I know, and I don't care the background. 🤣

PR #632, opened 2024-09-11
feat: [#631] implemented refOptions CLI argument

### LINKED ISSUE #631: Allow configuration of $refOptions in CLI application
Issue solved here: https://github.com/APIDevTools/json-schema-ref-parser/issues/199 introduced new argument in **json-schema-ref-parser**.
```javascript
$refOptions: {
        dereference: {
            externalReferenceResolution: "root"
        }
    }
```
This argument guarantees that during resolving deeper references (N-2+, references within references) the base URI will remain unchanged (which does not happen in default behaviour).

Currently it is not possible to pass `$refOptions` in **json-schema-to-typescript** CLI application, but it is already possible to configure it for `compile()` and `compileFromFile()` API methods. Due to this reason its not possible to use CLI application with schema layout as outlined in the linked issue to generate correct Typescript output.

I would like to ask for implementation of this option. Would do it myself, but I am not very experienced within JS/TS and with `minimist` library used to parse command line arguments. If full `$refOptions` are too much of an overhead to be added, I would be glad for at least option for `externalReferenceResolution` to be added separately and passed manually to the `$refOptions`.

Thanks in advance !

ndopj:
resolves: https://github.com/bcherny/json-schema-to-typescript/issues/631

Current changes accept `$refOptions` argument as JSON string from command line.

Those options are then being **merged** with default value for `$refOptions`. Even if current default value `{}` would be changed in the future, fields from command line argument `--refOptions` won't influence "not overridden" `$refOption` fields.

Command line argument `--refOptions` cannot use 1:1 naming with `Options` interface field `$refOptions` because it contains `$` character which might cause issues (on some Shells), if the command line arguments are not being enclosed as string (`""`).

**Still need to figure out if its possible to create any meaningful tests for this !!**
Any help or feedback appreciated.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -7,6 +7,8 @@ import isGlob from 'is-glob'
 import {join, resolve, dirname} from 'path'
 import {compile, DEFAULT_OPTIONS, Options} from './index'
 import {pathTransform, error, parseFileAsJSONSchema, justName} from './utils'
+import {ParserOptions as $RefOptions} from '@apidevtools/json-schema-ref-parser'
+import { merge } from 'lodash'
 
 main(
   minimist(process.argv.slice(2), {
@@ -26,7 +28,7 @@ main(
       'unreachableDefinitions',
     ],
     default: DEFAULT_OPTIONS,
-    string: ['bannerComment', 'cwd'],
+    string: ['bannerComment', 'cwd', 'refOptions'],
   }),
 )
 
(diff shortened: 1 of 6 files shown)


ndopj:
@bcherny Could I kindly ask for a review ? Don't know anybody amongst contributors so tagging you as an active one. I hope its not rude to you, in that case I am sorry :no_mouth:

Also if test should be included for this CLI option - I am not sure whether this would fit `testCLI.ts` structure. If I understand it correctly, tests are implemented via snapshots meaning it's asserted that output of each tested operation is same ? However, in such case, it would not be possible to include scenario described in the linked issue. I am correct ?

PR #634, opened 2024-09-16
fix invalid code generated due to empty `anyOf` array

k-yle:
`{ "anyOf": [] }` is currently converted to `type T = ()` which is invalid TypeScript syntax.

This PR fixes the logic so that empty sets and empty arrays become `never` instead of `()`.


k-yle:
hey @bcherny, sorry to ping you, could you please review this PR if you have time? 

It's a straightforward fix for a crash that makes this library unusable

Issue #627, opened 2024-09-02
Add option to prepend "readonly" to all fields


Bluscream:
This would be nice to ensure i don't accidentally try playing with values from responses but create my own local copy instead

Example: 
```
start "" json2ts -i "schemas/pietsmiet.de/*.schema.json" -o "src/responses/pietsmiet.de/" --inferStringEnumKeysFromValues true --readonly true
```

```ts
export interface PietsmietDEConfig {
    readonly API: Record<string, Record<string, string>>;
    readonly URLs: Record<string, string>;
    readonly Limits: Record<string, number>;
    readonly Orders: Record<string, string>;
    readonly Platform: Record<string, string>;
    readonly Stream: Record<string, string>;
  }
...
```

bcherny:
It would make sense to emit all types as readonly, gated behind a `--readonly` option.

PR #638, opened 2024-10-02
option to prepend "readonly" to all fields

Abhi-Bohora:
Linked Issue: #627

This PR introduces a new `readonly` option. When enabled, this option generates all properties in interfaces as readonly, including nested objects and arrays.


jonaskello:
I currently do this manually as a post-step, would be great if it could be built-in.

src/generator.ts
```diff
@@ -157,6 +157,10 @@ function generateTypeUnmemoized(ast: AST, options: Options): string {
     return `${type} | undefined`
   }
 
+  if (options.readonly && ast.type === 'ARRAY') {
```
bcherny, line 160: This doesn't seem to be covered by the tests. Mind adding coverage?

bcherny (changes requested):

PR #656, opened 2025-03-25
fix: escape enum values if they start with a number

n-buck:
Fixes the `SyntaxError: An identifier or keyword cannot immediately follow a numeric literal.` error,  if a enum value starts with a numbre.

### DIFF
--- a/src/generator.ts
+++ b/src/generator.ts
@@ -329,7 +329,7 @@ function generateComment(comment?: string, deprecated?: boolean): string {
 }
 
 function generateStandaloneEnum(ast: TEnum, options: Options): string {
-  const containsSpecialCharacters = (key: string): boolean => /[^a-zA-Z0-9_]/.test(key)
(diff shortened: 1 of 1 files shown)


bcherny:

This looks like it's already on `master` as of `1900e1f`: commit `a2234d3` ("Fix invalid TypeScript from inferStringEnumKeysFromValues", #694, merged 2026-08-01) rewrote the standalone-enum key-quoting check from a "contains special characters" regex to `isValidIdentifier` (`/^[a-zA-Z_$][a-zA-Z0-9_$]*$/`), which requires the first character not be a digit — exactly the case this PR fixes.

Repro built from your bug description (enum with a leading-digit member name via `tsEnumNames`), run against master:

```ts
export const enum LeadingDigitEnum {
  "1abc" = "1abc",
  normal = "normal"
}
```

No syntax error — the leading-digit key is quoted.

---

PR #670, opened 2025-09-22
Escaping

### LINKED ISSUE #171: Escape '/*' '*/' in comments
This is sometimes used in comments to refer to some Paths and make Prettier fail, resulting in a log message and an empty file.
It would be very handy to be able to escape those.

ex: `*/foo` becoming `*\/foo`

brettz9:
- fix: escape block comment by backslash instead of space; fixes #171
- chore: update lock file

The reason I updated the lock file was for some vulnerabilities fixed by `npm audit fix`.

~~The reason I updated ava was because its `engines` were not supporting Node 22 which I was using.~~

~~I added a `snapshots` script to `package.json` to update snapshots since this was not a transparent process and required for fixing changed test expectations.~~

The reason I changed the replacement from `* /` to `*\/` rather than making a normalizer rule per [this suggestion](https://github.com/bcherny/json-schema-to-typescript/issues/171#issuecomment-421885580) is because I think the default behavior should be to work as intended in TypeScript; namely, with the proper escaping, TypeScript will interpret the sequence as the termination of a comment block.

If you go to [this snippet](https://www.typescriptlang.org/play/?#code/PQKhCgAIUgBBTAHgQwLYAcA28o15AYwHsA7AZwBdJlIBeSAVgG599RIBlI1eQ7nklRAAdYK2iQAZgFcSBCgEtSkAEaQAFMhUEAlJADeAX1xiZcxcprq9RoA) and hover the mouse over the `function a () {}` block, you can see the comment is well-formed.


brettz9:
I've rebased.

bcherny:

This looks like it's already on `master` as of `d8618e5648c473bbf580a709684394d3dc2de523`: issue #171 (a `*/` inside a schema `description` prematurely closing the generated JSDoc comment) is already handled by `escapeBlockComment` in `src/utils.ts`, wired into the normalizer at `src/normalizer.ts:116-118`. Master replaces `*/` with `* /` (a space) rather than `*\/` (a backslash) as this PR proposes, but both approaches fix the same underlying bug — the generated comment no longer terminates early.

Repro against current master (`description: 'This is a /* nested */ comment'`):
```ts
export interface WithComment {
  /**
   * This is a /* nested * / comment
   */
  a?: string;
  [k: string]: unknown;
}
```
Compiles cleanly — no crash, no premature comment termination, no empty output.

(The bundled `package-lock.json` update for `npm audit fix` is a separate, now-stale change unrelated to the escaping fix, so I'm not weighing it here.)

Closing as superseded — but if you're still seeing the original bug on latest, say so and I'll reopen.



---

brettz9:
Thanks. I'm not seeing the bug, but I really think the better solution, the one which works with TypeScript without adding any spaces in the output as per https://github.com/bcherny/json-schema-to-typescript/issues/171#issuecomment-5160550288 is with the backslash instead of a space.

PR #672, opened 2025-10-08
fix(optimizer): simplify redundant intersections in allOf schemas

privatenumber:
## Problem

When a JSON Schema uses `allOf` with `$ref` that overlap, the generated TypeScript contains redundant intersections and duplicate properties.

Example from the [`tsconfig.json` schema](https://json.schemastore.org/tsconfig):

```ts
export type CompilerOptions = {
  types?: (string | null)[] | null;
  target?: string | null;
  [k: string]: unknown;
} & ({
  types?: (string | null)[] | null;  // Duplicate
  target?: string | null;             // Duplicate
  [k: string]: unknown;
} | null);
```

This happens when:

1. A schema definition is `$ref`’d multiple times
2. One `$ref` wraps the base in an `allOf` with extra constraints
3. The parser generates multiple AST nodes with the same name but different shapes

### Before

```ts
export type CompilerOptions = {...} & ({...} | null);
```

## Changes

1. Optimizer (`optimizer.ts`)

   - Simplifies `A & (A | null)` → `A | null`
   - Handles both `(A & (A | null))` and `(A | null) & A`
   - Deduplicates intersections like `A & B & A & B → A & B`

2. Generator (`generator.ts`)

   - Tracks generated type names with a `Set`
   - Skips duplicate `export type` declarations

### After

```ts
export type CompilerOptions = {...} | null;
```


test/__snapshots__/test/test.ts.md
```diff
@@ -444001,66 +443975,7 @@ Generated by [AVA](https://avajs.dev).
...
-      contentEncoding?: string;␊
-      if?: CoreSchemaMetaSchema;␊
-      then?: CoreSchemaMetaSchema;␊
-      else?: CoreSchemaMetaSchema;␊
-      allOf?: SchemaArray;␊
-      anyOf?: SchemaArray;␊
-      oneOf?: SchemaArray;␊
-      not?: CoreSchemaMetaSchema;␊
-      [k: string]: unknown;␊
-    } & (␊
```
privatenumber, line 444063: Looks like it was able to de-duplicate this too.

PR #673, opened 2025-10-25
feat: Add `removeOptionalIfDefaultExists` option

### LINKED ISSUE #558: Remove optional flag in case of `default` in schema.
Correct me if I'm wrong, but nowadays the converter follows the strict rule: make prop optional if it's not in `required` list.
But JSON Schema allows using key `default`. Of course, existence of the prop in real object depends on realization (e.g.: schema-based validator). 
I use `ajv` in my json-rpc server pipeline to validate arguments with `useDefaults:true`: it injects default values (defined in schema) in case of absent props.
That's why I'd like to have optional flag removed from such types.
If you feel this is good idea, I can try to fork and make a pull request soon. I think this behavior should be optional, e.g.: `removeOptionalIfDefaultExists"

sortA0329:
Fixes #558 

This PR adds a new option, `removeOptionalIfDefaultExists`, which allows the code generator to omit the optional modifier for properties that have a default value in the JSON Schema. The enhancement includes updates to the CLI, documentation, and core parsing logic, and is covered by new end-to-end tests.


calleli:
Perfect timing!
I realized this morning that I was looking for this functionality :)

Would love to have this in place!

yonatan2703:
When will this be merged and released?
We really want this feature

calleli:
> When will this be merged and released? We really want this feature

I forked the repo and brought in this functionality and then installed it using [patch-package](https://www.npmjs.com/package/patch-package). It does the trick for me in the meantime.

sortA0329:
@bcherny

I'd left this PR hanging since I no longer needed it, but it looks like there's more demand for it than I expected.
No rush, but if you get a chance, could you take a look?

Issue #363, opened 2021-01-14
Why does description expand referenced type?


SQReder:
Can I prevent reference type from expansion?

Input JSON:
```json
{
    "properties": {
      "correct": {
        "$ref": "#/definitions/color"
      },
      "wrong": {
        "description": "should be referenced",
        "$ref": "#/definitions/color"
      }

    },
    "additionalProperties": false,
    "definitions": {
        "color": {
            "type": "string",
            "enum": ["red", "green", "blue"]
        }
    }
}
```

Actual output:
```typescript
export type Color = "red" | "green" | "blue";

export interface Bug {
  correct?: Color;
  /**
   * should be referenced
   */
  wrong?: "red" | "green" | "blue";
}
```

Expected output:
```typescript
export type Color = "red" | "green" | "blue";

export interface Bug {
  correct?: Color;
  /**
   * should be referenced
   */
  wrong?: Color;
}
```

ericArbour:
I investigated this a bit it appears there's two things responsible for this behavior.

The first is that `json-schema-to-typescript` uses `json-schema-ref-parser` to dereference referenced definitions:
https://github.com/bcherny/json-schema-to-typescript/blob/8ea10a61f0d4b24749ae033a37f9aa3fb9a94417/src/resolver.ts#L11

From `json-schema-ref-parser`'s [docs](https://apitools.dev/json-schema-ref-parser/docs/ref-parser.html#dereferenceschema-options-callback): "The dereference method maintains object reference equality, meaning that all $ref pointers that point to the same object will be replaced with references to the same object."

And then `json-schema-to-typescript` uses this referential equality to determine if the types should be referenced or expanded:
https://github.com/bcherny/json-schema-to-typescript/blob/8ea10a61f0d4b24749ae033a37f9aa3fb9a94417/src/parser.ts#L123

But the presence of description on the referencing property breaks this referential equality because the referencing object is now a different object from the definition, and the result is the expanded type.

I'd love to work on a fix for this if the library maintainers have any advice on a good approach. I couldn't think of any solutions given the data available at the point where the referential equality check occurs.

johnbillion:
It's worth noting that this is not valid JSON schema in draft-07 and earlier. Support for siblings to a `$ref` were added in draft-09. https://json-schema.org/draft/2019-09/release-notes.html

booc0mtaco:
[...]

PR #683, opened 2026-05-11
Add pre-dereference normalization for annotation siblings in files

LudaThomas:
// JSON Schema 2019-09 allows annotation siblings next to `$ref`. When such
// siblings are present, json-schema-ref-parser produces a fresh merged object
// for the `$ref`, which breaks the referential equality the parser relies on
// to emit a named reference instead of inlining the target type. By wrapping
// the `$ref` in a single-member `allOf` before dereferencing, the dereferenced
// pointer keeps its identity and the annotations remain on the wrapper, where
// they can attach as a property-level JSDoc comment.

see #363


loyd:
@LudaThomas, I just want to say "thank you". Even if the PR won't be merged, it's clear how to modify the JSON schema to avoid this problem.

PR #648, opened 2024-12-18
Double quote standalone enum key if it contains special characters

troyshu:
Linked issue: https://github.com/bcherny/json-schema-to-typescript/issues/641

This PR ensures standalone enum keys with special characters are surrounded by double quotes, making the generated enums valid TypeScript.

PR #685, opened 2026-06-29
fix: escape enum member names so special characters produce valid output

mahirhir:
Object property keys go through `escapeKeyName`, which falls back to `JSON.stringify` for anything that isn't a plain identifier. Standalone enum member names don't get the same treatment: since #648 they're wrapped in double quotes when they contain special characters, but the contents aren't escaped. So a name with an embedded double quote (or a backslash, or a control character) is interpolated verbatim and breaks the output.

With `inferStringEnumKeysFromValues`, the schema

```json
{ "type": "string", "enum": ["a", "say \"hi\"", "b-c"] }
```

produces a member named `"say "hi""`, which isn't valid TypeScript. Because `format` is on by default, the prettier pass then throws `An enum member name must be followed by a ',', '=', or '}'` and `compile()` rejects. The same happens when the offending name comes from `tsEnumNames`.

The fix passes the name through `JSON.stringify` instead of building the quoted string by hand, which is what `escapeKeyName` already does for property keys. Output is unchanged for names that only contain non-quote special characters (`text/plain`, `b-c`, `d.e`), since e.g. `JSON.stringify("b-c") === '"b-c"'`, so the existing `enum.6` snapshot is untouched. `enum.7` covers a name with an embedded quote and now generates `'say "hi"'`.

This is orthogonal to #656: that one changes *when* a name is quoted (leading digit), this one changes *how* the quoted name is escaped. They compose.


bcherny:

This looks like it's already on `master` as of `1900e1f`: commit `a2234d3` ("Fix invalid TypeScript from inferStringEnumKeysFromValues", #694, merged 2026-08-01) — same commit that fixed #656. It changed standalone-enum key quoting to go through `JSON.stringify(keyName)` instead of raw string interpolation, exactly as this PR proposes.

Repro transcribed verbatim from your PR description (`enum: ['a', 'say "hi"', 'b-c']`, `inferStringEnumKeysFromValues: true`), run against master:

```ts
export const enum Enum {
  a = "a",
  'say "hi"' = 'say "hi"',
  "b-c" = "b-c"
}
```

Matches your `enum.7` snapshot exactly — `compile()` succeeds, no prettier/syntax error.

---

PR #697, opened 2026-08-02
Lazy-load prettier so format: false doesn't require node:v8

### LINKED ISSUE #679: Lazy-load prettier when format is set to false
Since prettier uses node specific modules i'm having issues running it in other runtime. Even if I turn off formatting with `format: false`, i still get noise in logs as following:

cloudflare workers:
```
[vitest-pool-workers] Failed to import "node:v8" from "/Users/shoeboom/Code/workspace/actions/node_modules/.pnpm/prettier@3.7.4/node_modules/prettier/index.mjs". To resolve this, try 
bundling the relevant dependency with Vite. For more details, refer to https://developers.cloudflare.com/workers/testing/vitest-
integration/known-issues/#module-resolution
```

I'd like to propose:
- lazy-load `prettier` only when `format: true`

Related issues:
- #189 discusses Prettier causing browser usage issues

Zelys-DFKH:
`src/formatter.ts` imported prettier at module scope, so requiring the package loaded prettier, and transitively `node:v8`, before any `compile()` call ran. That was true even with `format: false`. On a runtime that doesn't provide `node:v8` the import fails outright, which is what the report hit on Cloudflare Workers.

Thanks @ShoeBoom for the report and for pinning it to the Workers runtime.

The import now happens inside the `options.format` branch. `index.ts` only uses prettier's `Options` as a type, so that one became `import type`, which keeps a stricter module setting from quietly reintroducing the eager load.

The test needs two signals, which surprised me. Counting prettier entries in `require.cache` is the obvious check, and on its own it doesn't work: after this change prettier resolves through the ESM loader, and that loader never touches `require.cache`. The count stays at 0 even if someone later hoists the `import()` back out of the branch. So the test also checks whether `node:v8` got loaded, which catches the eager-`import` shape. I ran it against both regression shapes and it fails on each, then passes on this branch.

Verified on Node 16.20.2, 18.20.8, 20.20.2, 22.19.0 and 24.14.0, plus bun 1.3.14, on Windows and Linux. 173 tests pass, eslint and prettier are clean. `build:browser` is unaffected: browserify-shim maps prettier to a global there, and the downleveled dynamic import still goes through that shim, so the browser output gets lazier too.

Rebased onto master now that #696 has landed, and the test is ported to bun test. Details in the comment below.

Fixes #679


Zelys-DFKH:
Thanks for landing #696 first, it made this a nicer test to write. Rebased onto master and ported.

`test/test.ts` is gone, so the test now lives in `test/formatter.test.ts` and registers itself like the other suites. It shells out to `node` rather than `process.execPath`, the same way the CLI tests do. That part matters: under bun `process.execPath` is the bun binary, and bun doesn't populate `process.moduleLoadList`, which would have quietly made one of the two assertions vacuous.

Both assertions earn their keep. `require.cache` catches the import going back to module scope as a static import; `node:v8` catches it coming back as a hoisted `import()`, which resolves prettier's ESM entry and so never touches `require.cache`. I re-ran both regression shapes against the new runner and each one fails, each caught by a different signal.

Local run: 173 pass, `test/smoke.cjs` green on node 16, 18, 20, 22 and 24, and `bun dist/src/cli.js` still formats.

PR #701, opened 2026-08-02
Fix: don't infer tsEnumNames for enums normalized from const

### LINKED ISSUE #666: `const` is output as single value enum when `inferStringEnumKeysFromValues: true`
When transforming a schema with the `const` property and `inferStringEnumKeysFromValues: true`, the constant type is output as a single value enum. This can be an issue because it obfuscates the underlying type. It's specially troublesome when using `anyOf` with a constant field to disambiguate the object, such as in the second example.

<details>
<summary>Minimal Example</summary>

#### Schema
```json
{
	"type": "string",
	"const": "foo"
}
```

#### Expected Output
```ts
export type Root = 'foo';
```

#### Actual Output
```ts
export const enum Root {
	foo = 'foo'
}
```
</details>

<details>
<summary>Second Example</summary>

#### Schema
```json
{
	"type": "object",
	"anyOf": [
		{
			"properties": {
				"type": {
					"type": "string",
					"const": "a"
				},
				"foo": {
					"type": "number"
				}
			}
		},
		{
			"properties": {
				"type": {
					"type": "string",
					"const": "b"
				},
				"bar": {
					"type": "number"
				}
			}
		}
	]
}
```

#### Expected Output
```ts
export type Root =
	| {
			type?: 'a';
			foo?: number;
	  }
	| {
			type?: 'b';
			bar?: number;
	  };
```

#### Actual Output
```ts
export type Root =
	| {
			type?: Type;
			foo?: number;
	  }
	| {
			type?: Type1;
			bar?: number;
	  };

export const enum Type {
	a = 'a'
}
export const enum Type1 {
	a = 'b'
}
```
</details>

From what I understand, this is due to a [normalization](https://github.com/bcherny/json-schema-to-typescript/blob/master/src/normalizer.ts#L222-L227) where internally schemas with `const` are transformed to single value enum, likely for easier handling later? And because I have `inferStringEnumKeysFromValues` set to `true`, it automatically outputs the enum type for everything it thinks is an enum.

If this is something you agree should be fixed, I'd be willing to work on a PR, just let me know.

bcherny:
Fixes #666.

The normalizer rewrites `const` into a singleton `enum` internally, then the `inferStringEnumKeysFromValues` rule adds `tsEnumNames` to any string enum lacking them, making it indistinguishable from a real named enum. This tracks which schemas were rewritten from `const` and skips `tsEnumNames` inference for them, so a `const` schema keeps emitting as a plain string literal type instead of `export const enum`.

Does not change behavior for hand-written enums (single- or multi-value) — those still get `tsEnumNames` inferred as before.

Regression test: `test/e2e/options.inferStringEnumKeysFromValues.const.ts`.


bcherny:
---

PR #702, opened 2026-08-02
Fix `type: null` generating "any object" instead of `null`

### LINKED ISSUE #667: `type: null` generated as "any object" record
Schema:
```yaml
$schema: "https://json-schema.org/draft/2020-12/schema"
type: object
additionalProperties: false
required: [nullProp]
properties:
    nullProp:
        type: null
```
Generate with: `npx --package=json-schema-to-typescript@^15 json2ts -i schema.yaml -o schema.ts`
Produces (minus headers):
```typescript
export interface Schema {
  nullProp: {
    [k: string]: unknown;
  };
}
```
Expected:
```typescript
export interface Schema {
  nullProp: null;
}
```

Workaround: use `tsType`

bcherny:
Fixes #667.

`typesOfSchema()`'s `NULL` matcher only checked `schema.type === 'null'` (the string). A schema with `type: null` (the JS/JSON `null` value — an easy mistake, e.g. from an unquoted `null` in YAML) matched no type matcher at all, fell through to the `UNNAMED_SCHEMA` default, and `parseSchema()` rendered that as an "any object" `{[k: string]: unknown}` record. The fix makes the `NULL` matcher also accept `schema.type === null`, so it's treated the same as `type: "null"` and renders as `null`.

Does not cover other unrecognized `type` values (e.g. `type: "bogus"` or a schema with no `type` at all), which still fall through to the existing `UNNAMED_SCHEMA` behavior — verified unchanged.

Regression test: `test/e2e/nullType.ts`.


bcherny:
---

PR #703, opened 2026-08-02
Cap nested bounded-array tuple expansion by cumulative product

### LINKED ISSUE #690: Nested arrays with minItems/maxItems expand combinatorially and hang the compiler on default options
Found by fuzzing, reproduced by hand on `master` at `43ba08b` (Node v22.22.2, json-schema-to-typescript 15.0.3).

**Schema** (173 bytes, three nested bounded arrays of strings):

```json
{
  "type": "array",
  "items": {
    "type": "array",
    "items": {
      "type": "array",
      "items": {"type": "string"},
      "minItems": 1,
      "maxItems": 20
    },
    "minItems": 1,
    "maxItems": 20
  },
  "minItems": 1,
  "maxItems": 20
}
```

**Options:** defaults.

**What happens:** `compile()` never returns. I killed it at 60s, and again at 120s in an earlier run.

With `format: false` it does return, which shows where the work is going:

| `maxItems` (each level) | output |
| --- | --- |
| 18 | 40.6 MB |
| 19 | 55.6 MB |
| 20 | 75.0 MB |
| 21 | 99.8 MB |
| 22 | **0.00 MB** (falls back to unbounded arrays, returns in 0.2s) |

So a 173-byte schema generates 75 MB of TypeScript, and the hang on default options is Prettier trying to format it. Generation itself is fast (~0.2s) — the cost is the sheer volume.

**Expected:** either bounded output, or the same fallback to `T[][][]` that `maxItems: 22` already produces.

**Why it matters:** the cliff is invisible and inverted — `maxItems: 22` compiles instantly, `maxItems: 20` wedges the build. Nothing about the schema looks expensive, and 20 is exactly the default expansion limit, so the worst case sits right at the default. A bounded 20×20×20 grid is an ordinary thing to write.

**Root cause:** the guard added for this class of problem, `Remove maxItems if it is big enough to likely cause OOMs` (`src/normalizer.ts:150`), tests one array against `options.maxItems`:

```ts
if (maxItems !== undefined && maxItems - (minItems as number) > options.maxItems) {
  delete schema.maxItems
}
```

That is a per-array test, so each level here passes it (`20 - 1 = 19`, not `> 20`) while the levels multiply: ~21³ tuple combinations. It also explains the table above exactly — at 22, `22 - 1 = 21 > 20` trips the guard and the fallback kicks in.

**Possible direction:** carry the expansion already committed to by ancestor arrays and test the cumulative product against `options.maxItems`, rather than each array in isolation.

I have not opened a PR for this one. Changing when tuple expansion falls back is a semantics decision — it changes emitted types for schemas that compile fine today — so it seemed better left to you than guessed at. Happy to implement whichever direction you prefer.

bcherny:
Fixes #690.

The OOM guard in `normalizer.ts` checked each array's `maxItems` against `options.maxItems` in isolation, so nested bounded arrays (each within the limit on their own) could still multiply together into a combinatorial number of tuple variants. The fix walks up the parent chain and multiplies in each enclosing bounded array's fill count before comparing against `options.maxItems`, falling back to an unbounded array type at whichever nesting level would push the cumulative product over the limit.

Regression test: `test/e2e/arrayNestedMaxItems.ts`.


bcherny:
---

PR #704, opened 2026-08-02
Fix invalid index signature type when patternProperties/additionalProperties combine with named properties

### LINKED ISSUE #671: Override the key pattern for patternProperties ?
Would it be possible to add a setting to override TyeScript index key?

The following JSON schema

```json
{
  "type": "object",
  "title": "test",
  "properties": {
    "gid": {
      "type": "string"
    }
  },
  "patternProperties": {
    "^x-.*$": {
      "type": "number"
    }
  }
}
```

Returns a type that makes the named property unusable

```typescript
export interface Test {
  gid?: string;
  /**
   * This interface was referenced by `Test`'s JSON-Schema definition
   * via the `patternProperty` "^x-.*$".
   */
  [k: string]: number;
}
```

As `gid` will now be expected by TypeScript to be both a string and a number, even though my actual schema requires those properties to be prefixed with `x-` and they wouldn't conflict

I can work around this by setting `tsType: unknown` on the patterned property - at least TS will allow me to set `gid` to a string then. 

If I could to something like ``tsKey: "[K in `x_${string}`]"`` I could work around this without the project having to attempt to infer a proper TS template. It would also offer a workaround for #518

bcherny:
Fixes #671.

TypeScript requires every named property's type to be assignable to an interface's index signature type. The generated index signature was typed only from `patternProperties`/`additionalProperties`, ignoring sibling `properties`, so tsc rejected the output (TS2411). This widens the generated index signature type to also cover the named properties' types (including `| undefined` for optional ones), which is the root cause fix.

Does not cover overriding the pattern-property key pattern itself, which is a separate feature request and a design decision for the maintainer.

Regression test: `test/e2e/patternProperties.9.ts`.


bcherny:
---

PR #705, opened 2026-08-02
Fix justName truncating dotted $ref definition names

### LINKED ISSUE #645: `justName` with `$ref` definition may be unexpected
https://github.com/bcherny/json-schema-to-typescript/blob/25c1d2752ff4bad135d6fc53e80c2f99784b0e6f/src/normalizer.ts#L100

In our case, `dereferencedName` could be `#/definitions/v1.ManagedFieldsEntry`, it's not a filename but internal reference name, but `justName` consider it as `filename` always, it should consider `#` character actually. Currently, it returns `v1` but should be `v1.ManagedFieldsEntry` or `ManagedFieldsEntry`, right?

bcherny:
Fixes #645.

`justName` stripped everything after the first `.` in a dereferenced `$ref` (eg `#/definitions/v1.ManagedFieldsEntry` became `v1`), because it ran the JSON Pointer fragment through the same `stripExtension`/`extname` logic used for real file paths. The fix makes `justName` detect a `#` and, when present, use the last path segment of the fragment directly instead of extension-stripping it.

Does not cover file-extension stripping for real filenames (eg CLI/multi-file inputs) - that path is untouched.

Regression test: `test/e2e/refDottedDefinitionName.ts`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LbwECjWZ9dg1ERdrLGUdKH)_


bcherny:
---

PR #706, opened 2026-08-02
Fix unreachableDefinitions gating on declareExternallyReferenced

### LINKED ISSUE #652: unreachableDefinitions doesn't work without declareExternallyReferenced
I have many schema files, some of which contain multiple schemas through $defs

i.e.

```
{
  "$defs": {
    "CreateThingRequestBody": {
      "type": "object",
      "properties": {
        "title": {
          "type": "string"
        }
      }
    },

    "CreateThingResponseBody": {
      "properties": {
        "thing": {
          "$ref": "/schemas/entities/thing.schema.json"
        }
      },
      "additionalProperties": false
    }
  }
}
```

From the above file, I'd like to be able to generate a type for `CreateThingRequestBody` and `CreateThingResponseBody` without redeclaring the type for `Thing`

My script goes through my schemas file by file, so I want `Thing` to only be generated when it gets to `/schemas/entities/thing.schema.json`

`unreachableDefinitions` should achieve this, but for some reason it only works when `declareExternallyReferenced` is true

bcherny:
Fixes #652.

The generator's standalone-declaration gate only emitted a named type/interface when it was the root type or `declareExternallyReferenced` was set, which incorrectly also suppressed definitions pulled in solely via `unreachableDefinitions`. This tags ASTs produced from `unreachableDefinitions` so that flag independently satisfies the gate, matching the option's documented purpose.

Does not change any behavior for `$ref`-reachable types under `declareExternallyReferenced: false`, which are still not declared standalone.

Regression test: `test/e2e/options.unreachableDefinitions-declareExternallyReferenced-false.ts`.


bcherny:
---

PR #714, opened 2026-08-03
Fix unreachableDefinitions dropped when root schema isn't an object

### LINKED ISSUE #439: `unreachableDefinitions` not working with non plain object root node
If root node is not plain object type, `unreachableDefinitions` doesn't work and unreached definitions is not exported.

Calling: `json2ts -i test.schema.json -o test.d.ts --unreachableDefinitions`

Package version: 10.1.5
Node version: 17.3.1

## Number test
```json
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "type": "number",
  "definitions": {
    "unreached": {
      "type": "boolean"
    }
  }
}
```

### Received
```ts
export type TestSchema = number;
```

### Expected
```ts
export type TestSchema = number;
export type Unreached = boolean;
```

## Array test
```json
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "type": "array",
  "items": {
    "type": "number"
  },
  "definitions": {
    "unreached": {
      "type": "boolean"
    }
  }
}
```

### Received
```ts
export type TestSchema = unknown[];
```

### Expected
```ts
export type TestSchema = unknown[];
export type Unreached = boolean;
```

## Logic subschemas test
```json
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "type": "object",
  "oneOf": [
    {
      "type": "number"
    },
    {
      "type": "string"
    }
  ],
  "definitions": {
    "unreached": {
      "type": "boolean"
    }
  }
}
```

### Received
```ts
export type TestSchema = number | string;
```

### Expected
```ts
export type TestSchema = number | string;
export type Unreached = boolean;
```

## Object (only currently working)
```json
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "type": "object",
  "properties": {
    "test": {
      "type": "number"
    }
  },
  "definitions": {
    "unreached": {
      "type": "boolean"
    }
  }
}
```

### Received and expected
```ts
export type Unreached = boolean;

export interface TestSchema {
  test?: number;
  [k: string]: unknown;
}
```

bcherny:
Fixes #439.

The `unreachableDefinitions` handling lived inside `parseSchema`, which is only invoked for object-typed root schemas. This moves it to a new `parseUnreachableDefinitions`, called once from `compile()` on the root schema regardless of its type, with the resulting ASTs threaded through `optimize()`/`generate()` alongside the root AST.

Does not cover `$defs` nested inside non-root subschemas — those were already silently dropped before this fix (the definitions lookup used for naming only ever considered the root schema's own `$defs`), and remain out of scope here.

Regression test: `test/e2e/options.unreachableDefinitions-non-object-root.ts`.

auto-fix-bot:issue-439


percymcn:
I hit this same class of bug elsewhere today and was curious how it played out here, so I built the PR branch and measured it against its own parent rather than reasoning from the diff. Short version: **it works, it doesn't regress anything I could find, and its test genuinely discriminates.** Details below in case they're useful for review, since this is a bot-authored PR with no reviews yet.

**Before/after, same tree** — `d8618e5` (this PR's parent) vs `b857fe8` (PR head), `unreachableDefinitions: true`, one unreachable definition in each:

| root schema | parent | PR head |
|---|---|---|
| `type: "object"` | emitted | emitted |
| `type: "string"` | **dropped** | emitted |
| `type: "array"` | **dropped** | emitted |
| `anyOf: [...]` | **dropped** | emitted |
| no `type` at all | emitted | emitted |
| `type: "number"` + draft-07 `definitions` (the added fixture) | **dropped** | emitted |

That last row is worth calling out explicitly: **the new test fails on the parent commit**, so it's a real regression test rather than one that passes either way. That's not the norm for auto-generated fixtures and it's the first thing I'd have checked.

Both bag spellings work, which the diff explains — `parseUnreachableDefinitions` reads `rootSchema.$defs` off the *normalized* schema, so `definitions` has already been folded in by then.

**Things I tried to break and couldn't** (all byte-identical between parent and PR head):

- **Double emission.** `parseUnreachableDefinitions` parses *every* entry in `rootSchema.$defs`, including the reachable ones, so this was my main worry. It's fine — `processed`/`usedNames` are threaded through from the `parse()` call above it, so a reachable definition hits the cache. Worth knowing that the correctness depends on that ordering: if the two calls were ever swapped, reachable definitions would get `isUnreachableDefinition = true` and the "referenced by … via the `definition`" comment attached to them.
- **Reachable + unreachable in the same bag** — one of each, emitted once each.
- **Name collision with the root type name** (`{type: "string", $defs: {Root: {...}}}`) — correctly deduped to `Root1`. This one only becomes reachable *because* of the fix, since non-object roots emitted nothing before, so it seemed worth checking.
- Reachable definitions keep exactly the comment they had before.

`npm run format-check` and `npm run typecheck` both pass on the branch; I couldn't run the full suite because it needs `bun`.

**On the deferred half.** The description says `$defs` nested inside non-root subschemas were already dropped and stay out of scope — I verified both halves of that (dropped on the parent, still dropped on the PR head, so it's not a regression). For whatever it's worth in deciding whether that needs a follow-up: neither dominant generator produces that shape. `pydantic` 2.13.4 hoists every model to a root `$defs`, and `zod` 4.4.3 either inlines or hoists — I couldn't get either to emit a `$defs` at a non-root position. The population that would hit it is bundled schemas (2020-12 embedded resources keep their own `$defs`) and hand-authored/OpenAPI-style documents, which seems like a fair thing to defer.

**One small suggestion.** The fix's claim is "regardless of root type" and the fixture pins one root type (`number`). I verified `string`, `array`, and `anyOf` roots all work too, but none of them is pinned — and the immediately preceding commit (#708) was about combinator roots specifically, so an `anyOf` root is the one I'd add if you only add one. It's a two-line fixture.

PR #732, opened 2026-08-06
Emit JSON Schema defaults as JSDoc @default comments

aidinabedi:
## Summary

This PR adds support for emitting JSON Schema `default` values as TypeScript JSDoc `@default` comments.

Primitive defaults are added inline, while object and array defaults are formatted across multiple lines for readability.

Two new options provide ways to opt-out of the functionality:

| Option | Default | Description |
|---|---|---|
| `enableDefaultComments` | `true` | Set to `false` to suppress all `@default` comments |
| `prettifyDefaultComments` | `true` | Set to `false` to emit object/array defaults as single-line JSON |

## What changed

- emit `@default` JSDoc comments from schema `default`
- keep primitive defaults on one line
- format object and array defaults as multiline comments using Prettier

## Example

JSON Schema input:

```json
{
  "type": "object",
  "properties": {
    "enabled": {
      "type": "boolean",
      "default": true
    },
    "config": {
      "type": "object",
      "default": {
        "foo": "bar",
        "nested": {
          "1+1": 2,
          "hello": "world"
        }
      }
    }
  }
}
```

TypeScript output:

```ts
export interface Example {
  /**
   * @default true
   */
  enabled?: boolean;
  /**
   * @default {
   *   foo: "bar",
   *   nested: {
   *     "1+1": 2,
   *     hello: "world"
   *   }
   * }
   */
  config?: {
    [k: string]: unknown;
  };
}
```

## Tests

Added focused coverage for:
- all JSON primitives
- arrays
- nested objects
- multiline formatting behavior for complex defaults
- new options

The E2E test suite runs with `enableDefaultComments: false` to avoid regenerating a massive snapshot (which would otherwise results in +256,765 -256,449 changes). If you prefer to update the E2E tests, you can just revert or skip the last commit in this PR.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -18,8 +18,10 @@ main(
       'additionalProperties',
       'declareExternallyReferenced',
       'enableConstEnums',
+      'enableDefaultComments',
       'format',
       'ignoreMinAndMaxItems',
+      'prettifyDefaultComments',
       'strictIndexSignatures',
       'unknownAny',
       'unreachableDefinitions',
@@ -181,6 +183,8 @@ Boolean values can be set to false using the 'no-' prefix.
       Declare external schemas referenced via '$ref'?
   --enableConstEnums
       Prepend enums with 'const'?
+  --enableDefaultComments
+      Generate @default JSDoc comments for properties with a default value
   --inferStringEnumKeysFromValues
       Create enums from JSON enums instead of union types
   --format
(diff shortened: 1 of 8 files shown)


aidinabedi:
@bcherny If you want, I can add a way to disable defaults in Options. And then disable it in e2e tests, so this PR doesn't need to update snapshot `e2e.test.ts.snap` (which results in +256,765 -256,449 changes)

aidinabedi:
I decided to add two compile options to give users more control over the `@default` comment behavior and to opt-out of updating snapshot e2e.test.ts.snap in this PR

PR #739, opened 2026-08-10
Fix validator traversal of if/then/else keywords (#626)

### LINKED ISSUE #626: The key property named deprecated is wrong behavior
I Cannot generate OpenAPI v3.1.0 types from json schema.

https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json

```bash
> curl https://spec.openapis.org/oas/3.1/schema/2022-10-07 | pnpm json2ts > ./src/openapi/OpenAPIV31.d.ts
[
  'Error at key "properties" in file "undefined": deprecated must be a boolean'
]
[
  'Error at key "properties" in file "undefined": deprecated must be a boolean'
]
[
  ValidationError
      at C:\Users\user\Documents\workspace\hono-typia-openapi\node_modules\.pnpm\json-schema-to-typescript@15.0.1\node_modules\json-schema-to-typescript\dist\src\index.js:97:19
      at Generator.next (<anonymous>)
      at fulfilled (C:\Users\user\Documents\workspace\hono-typia-openapi\node_modules\.pnpm\json-schema-to-typescript@15.0.1\node_modules\json-schema-to-typescript\dist\src\index.js:5:58)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
]
```

joyheroes:
## Summary

Fixes a validator traversal bug where JSON Schema `if`, `then`, and `else` keywords were incorrectly treated as schema maps instead of individual schemas.

This caused valid schemas containing properties named `deprecated` inside conditional branches to fail validation with:

`deprecated must be a boolean`

Fixes #626.

## Root cause

`traverse()` had explicit handling for some schema-valued keywords, but not for `if`, `then`, and `else`.

Their values were incorrectly processed by the generic object traversal path, causing nested `properties` objects to be visited at the wrong level.

## Changes

- Add explicit traversal handling for `if`, `then`, and `else`.
- Add these keywords to `BLACKLISTED_KEYS`.
- Add regression coverage for conditional schemas containing `deprecated` properties.

## Testing

Verified:

- format check
- lint
- typecheck
- full test suite

Result:

288 tests passed.

Snapshots:

155 passed, 1 added.

## Scope

This PR only fixes the validator false positive described in #626.

It does not address unrelated OpenAPI 3.1 recursive reference limitations.


joyheroes:
Hi — I rechecked this against current master and the regression still reproduces there: `traverse()` in `src/utils.ts` still has no explicit `if`/`then`/`else` handling. I also reran the project suite on this branch (`npm test`: 288 passed, 156 snapshots) and a reverse-validation that temporarily reverts the fix reproduces the `deprecated must be a boolean` failure on the `ifThenElse` case. Happy to adjust the approach if you'd prefer this handled differently.

